PR #1, opened 2016-09-05
modification to be bbox and route type generic

jamescr:
Little modifications in code just to be able to work with a different bbox and route type (passed as parameters in some methods). The default usage of the code should work nice for Florianopolis bus routes. 

Note: This modifications are being used in the incofer2gtfs.py script of the forked osm2gtfs.

### DIFF
--- a/osm2gtfs.py
+++ b/osm2gtfs.py
@@ -10,6 +10,8 @@ from datetime import datetime
 from osmhelper.osm_routes import Route, RouteMaster
 
 
+FLORIANOPOLIS = {"e":"-48.27117919921875", "n":"-27.215556209029675", "s":"-27.94103350326715", "w":"-49.0155029296875"}
+
 DEBUG_ROUTE = "104"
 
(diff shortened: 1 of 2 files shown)


grote:
Awesome, thanks a lot! :smile:  I think small incremental modifications like these are the way to go!

Maybe here we can even go a step further. I don't really like any location specific information in the `osmhelper`. So how about moving the `FLORIANOPOLIS` bbox to the `osm2gtfs.py` for the time being and not have any default bboxes in the get* methods?

jamescr:
I agree with moving the specific info to osm2gtfs.py and having no defaults for bbox, but right now to achieve that, the position of the "refresh" arg of get* methods must be moved to the end, otherwise there will be a "SyntaxError: non-default argument follows default argument". If you don't worry about those args reordering, I'll commit them.

grote:
Thanks a lot. I had to fix a couple of issues, but merged your PR now.

PR #8, opened 2016-09-17
Added requirements.txt, configuration into separate file and code cleanup

xamanu:
The idea behind all work here is to get osm2gfts to become truly generic!

This means, I don't want to break the current functionality to export Florianópolis at any time. But to improve and extend osm2gtfs to enable us to generate the GTFS of Managua from the same code base. All steps are in the direction of getting things abstracted in order to integrate our own functionality.

This effort should also allow @jamescr similarly to integrate his way of importing schedule data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,10 @@ to create a GTFS file using Google's transitfeed library and validates it after 
 
     pip install -r requirements.txt
 
+## Use
+
+    python osm2gtfs.py -c fenix.json.example
+
 ## License
 
 ![GNU GPLv3 Image](https://www.gnu.org/graphics/gplv3-127x51.png)
--- a/README.md
+++ b/README.md
@@ -6,8 +6,7 @@ Use public transport data from OpenStreetMap and external schedule information t
 **Attention:** The source code is currently very specific to one city, but this can be developed into something more generic.
 You can help with pull requests that introduce an abstraction layer for the external schedule information.
 
-How does it work?
------------------
+## How does it work?
 
 The osmhelper is responsible for retrieving current data directly from OpenStreetMap
 via the Overpass and OSM API.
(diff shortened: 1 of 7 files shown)


xamanu:
The idea behind all work here is to get osm2gfts to become truly generic!

This means, I don't want to break the current functionality to export Florianapolis at any time. But to improve and extend osm2gtfs to enable us to generate the GTFS of Managua from the same code base. All steps are in the direction of getting things abstracted in order to integrate our own functionality.

This effort should also allow @jamescr similarly to integrate his way of importing schedule data.

xamanu:
It'd be useful to get the data from Florianapolis, so I can run the script with the original data and make sure it always works.

README.md
```diff
@@ -6,23 +6,29 @@ Use public transport data from OpenStreetMap and external schedule information t
...
-* https://github.com/metaodi/osmapi
+## Depends
 
-License
--------
+- https://github.com/DinoTools/python-overpy
+- https://github.com/google/transitfeed
+- https://github.com/metaodi/osmapi
+
+## Install
```
grote: This point isn't really about the install of the script, but rather about the installation of the dependencies. Maybe it could be clarified like this?

fenix.json.example
```diff
@@ -0,0 +1,25 @@
...
+        "agency_lang": "pt",
+        "agency_phone": "+55 (48) 3025-6868"
+    },
+    "feed_info": {
+        "publisher_name": "Torsten Grote",
+        "publisher_url": "https://transportr.grobox.de",
+        "version":  "0.1"
+    },
+    "start_date": "20160901",
+    "end_date": "20170831"
```
grote: For the dates, we need a more flexible solution, because when used with Navitia, the difference between start and end date can never be more than a year and ideally, for this use-case, I like these dates to automatically move forward when I run the script again without needing to update them somewhere manually.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+overpy==0.3.1
+transitfeed==1.2.14
+osmapi==0.6.2
```
grote: Is there any reason to require exact versions here?
I'd say we replace the `==` with `>=` because it is very likely that things will continue to work with future versions of these libraries.

fenix.json.example
xamanu: Good idea, but I opened an issue (#14) for this, because i't's a feature request. Right now we first need to get the functionality which is already there and make the script useful for others/us. New features shouldn't be show stoppers in this process.

grote: That's fine with me, but then these dates should be moved into the `feed_info` section.

requirements.txt
xamanu: No reason. Probably we should just omit any version specification.

README.md
xamanu: > Install dependencies

Would this be a suitable title for you?

grote: Yes, that'd be fine.

requirements.txt
grote: Having a minimum version to start from is not a bad idea, because we never tested very old versions and they might fail miserable. Using `>=` at least starts with the latest known to work version.

Issue #3, opened 2016-09-17
Abstract configuration


xamanu:
Hello,

currently a lot of custom configuration data is stored in code. I'd like to push it to be more generic.

I'd suggest to rely on a config.json file similar to geom2gtfs does: https://github.com/conveyal/geom2gtfs/blob/master/kingco.json.example

We could start with a simple structure, based on the general information used for querying from OSM and from GTFS' agency.txt and start and end date from calendar.txt:

	{
		"query":{
	        	"network": "", 
	        	"bbox": {"e": "-48.2711", "n": "-27.2155", "s": "-27.9410", "w": "-49.0155"},
	   	},
		"agency":{
	        	"agency_id": "",
	        	"agency_name": "", 
	        	"agency_url": "", 
	        	"agency_timezone": "",
	        	"agency_lang": "",
	        	"agency_phone": "", 
	        	"agency_fare_url": "",
	   	},
		"feed_info":{
	        	"publisher_name": "", 
	        	"publisher_url":  "",
	        	"licence_name":  "",
	        	"licence_url":  "",
	        	"version":  "",
	   	},

		"gtfs_mode":3,
		"start_date": "",
		"end_date":"",
		"schedule_information":  "",
		"output": "", 
	}

What do you think? Is this a way to go?

grote:
Thanks for creating this ticket. We should definitely get as much specific information out of the code as possible and we should do it in small incremental steps.

@jamescr will like the json format above as that looks very similar to what he is already using [in his fork](https://github.com/jamescr/osm2gtfs/blob/master/data/input_incofer.json).

How about we start with extracting the `feed_info` and `agency` as a first step and then create additional tickets for the other data that *could* also live in this same json file?

xamanu:
Sounds good!

xamanu:
I put the basic configuration into a separate json config file. It's in the recent pull request. I appreciate your consideration!

grote:
> I put the basic configuration into a separate json config file. It's in the recent pull request. I appreciate your consideration!

Thanks a lot! Btw., there's no need to mention this here again. I get notified about new PRs and changes to them ;)

jamescr:
:+1: for using a json file for the config data., I'll comment on the PR.

PR #16, opened 2016-09-18
#3 - Moved basic configuration into separate json file

xamanu:


osm2gtfs.py
```diff
@@ -31,18 +27,32 @@
 
 
 def main():
+
+    # Load config json file
+    try:
+        config = json.load(args.config)
```
grote: The error handling isn't really working here when the script is called without parameters:
```
  File "./osm2gtfs.py", line 33, in main
    config = json.load(args.config)
  File "/usr/lib64/python2.7/json/__init__.py", line 287, in load
    return loads(fp.read(),
AttributeError: 'NoneType' object has no attribute 'read'
```
I think it should take `config.json` by default and show a proper error message when the file can not be found.

fenix.json.example
```diff
@@ -0,0 +1,25 @@
+{
```
grote: Maybe this file's name should be `fenix.json` instead. But since it is the first config file in the repo, it could also be called `config.json.example`. Then there should be a a line in the readme telling people to copy this file to `config.json` (which should be added to `.gitignore`).

fenix.json.example
```diff
@@ -0,0 +1,25 @@
...
+        "agency_lang": "pt",
+        "agency_phone": "+55 (48) 3025-6868"
+    },
+    "feed_info": {
+        "publisher_name": "Torsten Grote",
+        "publisher_url": "https://transportr.grobox.de",
+        "version":  "0.1"
+    },
+    "start_date": "20160901",
+    "end_date": "20170831"
```
grote: Since these dates will be added to the `feed_info` file, I think it should also be moved into this dictionary.

osmhelper/__init__.py
```diff
@@ -286,7 +286,7 @@ def refresh_route(route_ref, route_type, bbox):
         <bbox-query e="%s" n="%s" s="%s" w="%s"/>
     </query>
     <print/>
-    """ % (route_type, str(route_ref), bbox["e"], bbox["n"], bbox["s"], bbox["w"]))
+    """ % (route_type, str(route_ref), str(bbox["e"]), str(bbox["n"]), str(bbox["s"]), str(bbox["w"])))
```
grote: You have those as strings in the json file, why is it necessary to cast them to strings here again?

osm2gtfs.py
```diff
@@ -31,18 +27,32 @@
...
+    except ValueError, e:
+        print('Failed to load config json file:')
+        print(e)
+        sys.exit(0)
+
+    # Initialize information from config fimle
```
grote: There's a typo: `fimle`

osm2gtfs.py
xamanu: Argparser does this for us. Just not in the group, which did make sense at all. Now it should be fine.

fenix.json.example
xamanu: Done.

fenix.json.example
xamanu: I like it as it is. It explains it is fenix not just another plain config.json.example and fenix.json would let me guess it is something which is useful for everybody, which isn't

osmhelper/__init__.py
xamanu: Because we are not using Python 3 and json data structures are imported and stated as unicode. Without those it just doesn't work :)

osm2gtfs.py
grote: Making the config file argument required is not the fix I meant. I was suggesting to have `config.json` as a default argument for the config parameter, but don't include this file in the repository and even add it to `.gitignore`.

This enables people to have their local config file and play with it without it being tracked by version control and without having to specify the file every time they want to run the script. If people are happy with the official file in the repository, they can symlink `config-fenix.json` to `config.json`.

fenix.json.example
grote: how about `config-fenix.json` then to have all different config files with the same prefix, so they can be found easily and are grouped together in a directory listing?

osm2gtfs.py
```diff
@@ -31,18 +27,32 @@
...
+    # Load config json file
+    try:
+        config = json.load(args.config)
+    except ValueError, e:
+        print('Failed to load config json file:')
+        print(e)
+        sys.exit(0)
+
+    # Initialize information from config file
+    bbox = config['query']['bbox'];
```
grote: There's a semicolon that shouldn't be there.

fenix.json.example
grote: An example is not really needed if we have working configs in the repo.

osm2gtfs.py
```diff
@@ -82,28 +92,28 @@ def main():
 
     schedule = transitfeed.Schedule()
     agency = schedule.AddAgency(
```
jamescr: Another approach is in my fork, I which I add a method that returns a *transitfeed.Agency* object created using info from a json file (input_incofer.json), in the json file the valid keys are any of the *transitfeed.Agency._FIELD_NAMES* array values.

osm2gtfs.py
xamanu: Fixed

osm2gtfs.py
xamanu: Fixed

osm2gtfs.py
xamanu: I like this a lot! For now (in this pull request), I just introduced the config file which also has the same data structure. So, we can definitively use this call in the code to generate the agency gtfs information nicely.

fenix.json.example
xamanu: Can we handle this in a separate issue, please? I think as it is proposed here, it's already an improvement.

osm2gtfs.py
xamanu: Can we handle this in a separate issue, please? I think as it is proposed here, it's already an improvement.

grote: Alright, I created #18 for this.

osm2gtfs.py
jamescr: I create a separate issue (#21)  for that.

Issue #18, opened 2016-09-19
Don't require config file parameter to run script


grote:
Making the config file argument required is not ideal. I suggest to have `config.json` as a default argument for the config parameter, but don't include this file in the repository and even add it to `.gitignore`.

This enables people to have their local config file and play with it without it being tracked by version control and without having to specify the file every time they want to run the script. If people are happy with the official file in the repository, they can symlink `config-fenix.json` to `config.json`.

PR #20, opened 2016-09-19
#18 - Introduced standard config.file location

xamanu:


osm2gtfs.py
```diff
@@ -29,11 +30,13 @@
...
-    try:
-        config = json.load(args.config)
-    except ValueError, e:
-        print('Failed to load config json file:')
-        print(e)
+    if os.path.isfile('config.json'):
+        with open("config.json") as json_file:
+            config = load_config(json_file)
+    elif args.config is not None:
+            config = load_config(args.config)
```
grote: What happens here if the config file as been specified as an argument, but does not exist?

grote:
Thanks for this PR! :)

Unfortunately, I can't merge it right away because it conflicts with the other one I just merged.

osm2gtfs.py
xamanu: > usage: osm2gtfs [-h] [--config FILE] [--output FILENAME]
                [--refresh-route ROUTE | --refresh-all-routes | --refresh-all-stops | --refresh-all]
osm2gtfs: error: argument --config/-c: can't open 'foo.json': [Errno 2] No such file or directory: 'foo.json'

xamanu:
I re-rolled the commit. Hopefully it works now. Thanks for feedback and pulling in.

PR #22, opened 2016-09-19
fix issue#21

jamescr:


osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
...
+def load_agency(config):
+    '''
+    Loads agency data from a json config file. The valid keys under the json
+    "agency" object correspond to the transitfeed.Agency._FIELD_NAMES.
+
+    return a transitfeed.Agency object
+'''
+
+    data_dict = {}
+    for field_name in transitfeed.Agency._FIELD_NAMES:
```
grote: This variable is protected, we should not access it directly. I'll change this before merging.

osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
 
     return config;
 
+def load_agency(config):
```
grote: Just a German style nitpick for the future ;) PEP8 requires two blank lines between methods.

osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
...
+    '''
```
grote: Great you are adding documentation already! Docstrings should be made with double quotes `"` btw.

osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
...
+    data_dict = {}
+    for field_name in transitfeed.Agency._FIELD_NAMES:
+        try:
+
+            field_value = config["agency"][field_name]
+            if field_value is not None and field_value != "":
+                data_dict[field_name] = field_value
+
+        except KeyError:
+            sys.stderr.write("key '"+field_name+"' was not found on the config file.")
```
grote: I think it should be "*in* the config file".

osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
...
+
+    data_dict = {}
+    for field_name in transitfeed.Agency._FIELD_NAMES:
+        try:
+
+            field_value = config["agency"][field_name]
+            if field_value is not None and field_value != "":
+                data_dict[field_name] = field_value
+
+        except KeyError:
```
grote: I think it would be better to check if `field_name is in config["agency"]` rather than catching `KeyError`s.

osm2gtfs.py
```diff
@@ -103,12 +103,7 @@ def main():
     stops = osmhelper.get_stops(routes)
 
     schedule = transitfeed.Schedule()
-    agency = schedule.AddAgency(
-        name=config['agency']['agency_name'],
-        url=config['agency']['agency_url'],
-        timezone=config['agency']['agency_timezone'],
-        agency_id=config['agency']['agency_id']
-    )
+    agency = load_agency(config)
```
grote: This will break the script, because the agency is never added to the schedule object. Did you test this?

osm2gtfs.py
```diff
@@ -298,5 +293,34 @@ def load_config(file):
...
+
+        except KeyError:
+            sys.stderr.write("key '"+field_name+"' was not found on the config file.")
+
+    agency = transitfeed.Agency(field_dict=data_dict)
+
+    if not agency.Validate():
+        sys.stderr.write("Agency data has errors")
+        # later, an error handling based on a exception here could be added.
+        #raise AttributeError('Agency data on config file in not valid.')
```
grote: Maybe we should distinguish here between required and optional fields. Do you know how the validation behaves? Maybe it prints warnings already?

osm2gtfs.py
grote: Sometime later, we should maybe distinguish here between required and optional fields or does the validation already print warnings for that?

osm2gtfs.py
jamescr: You are right, I'm sorry. I thought I tested but obviously I make a mistake.

osm2gtfs.py
jamescr: The validate method returns a boolean and have a mechanism to handle problems. I think the method "load_agency" could be improved to be less naive.

jamescr:
Maybe could be more simple,  just invoke the Agency constructor like this: 

agency = transitfeed.Agency(field_dict=config["agency"])

there is no more need of the load_agency method.

Issue #9, opened 2016-09-17
Abstract schedule data input


xamanu:
Currently, the data about the schedule information is getting pulled in a very custom way into the script. 

Others would like to use the script. For now I see different ways for feeding the script:

- Currently @grote uses a json file with route numbers and then pulls data from the Fenix network website and parses the time information.
- @jamescr is putting schedule information directly in a json file
- We (MapaNica) would like to use a CSV file with frequencies and run some logic on it.

I suggest moving different ways on pulling schedule data into different files, which could be handled similar to small plugins. Basically all of them then have to output the same information to the main script: Stop times of the first and last stops (mostly terminals, I guess).

Probably we could specify in the config file (#3) the name of the schedule information plugin. Do you have any suggestions on how to implement this properly?

This is an important step in order to make this tool generically usable.

grote:
I think this should be the last step of making this script truly general, also because it is the most difficult one and probably the one where we are not yet sure what we really need and how these yet unknown use-cases can be covered best by an abstraction layer.

Just feeding in static files might be to inflexible. I expect most people still needing to integrate some logic. There could maybe be something like a abstract `GTFSWriter` class that already has methods for most tasks and is essentially a wrapper around the `transitfeed` library. This class could then be extended by different region specific classes that need to overwrite some functionality.

But that's just a first idea. Let's see how exactly the various use-cases look like and what architecture would accommodate them best.

xamanu:
_Disclaimer: I understand, that you want to go slow and do everything very well. But this task here is definitively the most **important** step for us and as you did by writing a custom script, we also need a solution as soon as possible and it cannot be the last step. I'd consider all progress that is better than the current status an improvement._

In a first, more or less easy and working step, I was thinking about an abstract `ScheduleImporter` class which then inherits to sub classes, such as `FenixImporter`, `IncoferImporter`, `MapanicaImporter`. I'd lean very closely to the current implementation and basically let these classes to import and do logic on data in the constructor and then have methods to serve schedule data for weekdays, saturday and sunday.

grote:
If you like, I don't mind if you start factoring out things into different classes. You should just be prepared that this *might* need to be refactored again. Maybe it would also be a good idea to start specifying the classed (maybe in UML?) here in this issue before the implementation is started, so we have a change to discuss the architecture first.

xamanu:
[...]

PR #25, opened 2016-10-01
#9 - Abstracting schedule creating structure; Introducing factory

xamanu:
This is a bigger pull request and I guess it needs some testing before we can merge this in. But it's definitively going to be a mayor step towards integration of other cities into this script. With this @jamescr and myself should be able to use the code base for our cities to generate GTFS :dancers: 

**Changelog**

- The main abstraction is happening through the `CreatorsFactory`. It initiates creator objects for each parts of the GTFS feed (agency, feed_info, routes, stops and trips). Those use a standard implementation and can be easily overwritten by naming convention based on the `selector` value from the config file.
- All existing code has been moved into respective creator classes, without touching any of the existing logic.
- All default implementations live in the `creators_default` folder and all region specific code lives in `creators_custom` folder.
- Right now there are _working default implementations_ for agency and feed_info because these were already (somehow) generic based on values from the config file.
- Right now there are still _no default implementations_ for routes, stops and trips as their current code exclusively works for the Fenix (Florianópolis) network. The default creator classes only define the basic functions that those creator are required to implement.


xamanu:
I was not able to test it properly with the Fenix network because the script doesn't run through (even without the changes) because of the quantity of overpass queries it contains. I was playing around with the bounding box and I could do some testing, but it's likely to still have some issues.

xamanu:
BTW, this looks like a huuuge pull request, but most of it, is moving existing code into classes that follow a different but easy structure which is abstractable and allows others to inject their own code and logic. This was necessary but also consists in probably the smallest step possible to allow future pull requests to keep following out step by step pattern.

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,501 @@
...
+        if isinstance(member, overpy.RelationNode):
+            stop_type = "node"
+        elif isinstance(member, overpy.RelationWay):
+            stop_type = "way"
+        else:
+            raise RuntimeError("Unknown type: " + str(member))
+
+        return Stop(stop_id, stop_type)
+
+class BaseRoute(object):
```
grote: Do you plan to have these classes as part of the route creator or is this just the first step to keep the PR small?

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,501 @@
...
+
+    def add_routes_to_schedule(self, schedule):
+
+        routes = creators_custom.ROUTES
+
+        blacklist = ['10200', '12400', '328', '466', '665']
+        # Try to find OSM routes in Fenix data
+        for route_ref, route in sorted(routes.iteritems()):
+            found = False
+            for linha in linhas:
```
grote:
```
  File "./osm2gtfs.py", line 103, in <module>
    main()
  File "./osm2gtfs.py", line 76, in main
    routes_creator.add_routes_to_schedule(schedule)
  File "/home/user/vcs/osm2gtfs/creators_custom/RoutesCreatorFenix.py", line 35, in add_routes_to_schedule
    for linha in linhas:
NameError: global name 'linhas' is not defined
```

creators_custom/StopsCreatorFenix.py
```diff
@@ -138,25 +217,6 @@ def get_interim_stop_name(self):
...
-            raise RuntimeError("Unknown type: " + str(member))
-
-        return Stop(stop_id, stop_type)
-
-    @staticmethod
-    def normalize_name(old_name):
-        name = STOP_REGEX.sub(r'\1', old_name)
-        name = name.replace('Terminal de Integração da Lagoa da Conceição', 'TILAG')
-        return name
-
```
grote: Where did these methods end up?

Maybe we can have default OsmRoute and OsmStop classes that don't need to be overwritten and these custom methods can live in the `StopCreator` and work on a OsmStop object?

creators_custom/RoutesCreatorFenix.py
xamanu: Yes, in order to keep this PR as small as possible, II was not changing any of the logic or finer structure. Things like these are going to be handled in #26.

creators_custom/StopsCreatorFenix.py
xamanu: This method was moved in the Route class because it gets only called from there. This was easier than doing a lot of cross imports of classes that otherwise do not interfere.

creators_custom/RoutesCreatorFenix.py
xamanu: A good example what would be different in Fenix from a default implementation (#26) :smile:
I moved the linhas.json (with schedule information) from the TripsCreator to the RoutesCreator. Sorry, it's very anoying I can't test this myself because of constant errors:
`overpy.exception.OverpassTooManyRequests: Too many requests`

creators_custom/RoutesCreatorFenix.py
grote: Awesome! :)

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,508 @@
...
+from osmapi import OsmApi
+from creators_default.RoutesCreator import RoutesCreator
+import creators_custom
+
+CHECK_FOR_NON_PLATFORMS = False
+
+class RoutesCreatorFenix(RoutesCreator):
+
+    def __init__(self, config):
+        self.config = config
```
grote: I think it would be better to call the super constructor here which already does assign the config.

creators_custom/RoutesCreatorFenix.py
grote: Sorry for the inefficient API use :(

I sent you an email, so you can test this better, because doing this blind is not going to work ;)
There's a couple of other issues in this class as well.

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,508 @@
...
+                print ",",
+            i += 1
+        print "];"
+        i = 0
+        for node in self.shape:
+            print "L.marker([" + str(node["lat"]) + ", " + str(node["lon"]) + "]).addTo(map)"
+            print "    .bindPopup(\"" + str(i) + "\").openPopup();"
+            i += 1
+
+    def normalize_stop_name(old_name):
```
grote: I think this can again be a static method, otherwise it needs `self` as first parameter.

xamanu: Yes, thanks. Fixed.

creators_custom/RoutesCreatorFenix.py
xamanu: Yes, this is true and makes a lot of sense. Fixed.

creators_custom/RoutesCreatorFenix.py
grote: Hmm yes they were created with the old structure. I can't create new ones due to the issues in this file. Will highlight the problems with comments :(

xamanu: Bäh. Sorry for that. I'm wondering why you don't get the _OverpassTooManyRequests_ exception and I do :unamused:

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,499 @@
...
+                sys.stderr.write("    https://www.openstreetmap.org/relation/" + str(rel.id) + "\n")
+
+            return routes[ref]
+        else:
+            sys.stderr.write("Route has no ref tag: " + "https://www.openstreetmap.org/relation/" + str(rel.id) + "\n")
+
+
+    def refresh_route(self, route_ref, route_type):
+
+        route_masters = get_route_masters(route_type)
```
grote: unresolved reference to `get_route_masters`

and too many blank lines (PEP30)

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,499 @@
...
+            # TODO also refresh route master
+            print route_masters[route_ref]
+        elif route_ref in routes:
+            print str(routes[route_ref])
+        elif len(result.get_relations()) == 0:
+            sys.stderr.write("No route %s found" % route_ref)
+            return
+
+        for rel in result.get_relations():
+            print get_route(routes, route_masters, rel, warn=False)
```
grote: unresolved reference `get_route`

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,499 @@
...
+
+        save_routes_to_file(routes)
```
grote: unresolved reference `save_routes_to_file`

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,499 @@
...
+        stop_id = member.ref
+
+        if isinstance(member, overpy.RelationNode):
+            stop_type = "node"
+        elif isinstance(member, overpy.RelationWay):
+            stop_type = "way"
+        else:
+            raise RuntimeError("Unknown type: " + str(member))
+
+        return Stop(stop_id, stop_type)
```
grote: unresolved reference `Stop`

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,499 @@
...
+        print "];"
+        i = 0
+        for node in self.shape:
+            print "L.marker([" + str(node["lat"]) + ", " + str(node["lon"]) + "]).addTo(map)"
+            print "    .bindPopup(\"" + str(i) + "\").openPopup();"
+            i += 1
+
+    @staticmethod
+    def normalize_stop_name(old_name):
+        stop_regex = re.compile('(TICAN|TISAN|TICEN|TITRI|TILAG|TIRIO|TISAC).*')
```
grote: unresolved reference to `re`

creators_custom/StopsCreatorFenix.py
```diff
@@ -1,14 +1,84 @@
...
+                print stop
+
+        # save stops
+        with open('data/stops.pkl', 'wb') as f:
+            pickle.dump(stops, f, pickle.HIGHEST_PROTOCOL)
+
+        return stops
+
+    def fill_stops(stops, route):
+        if isinstance(route, Route):
```
grote: unresolved reference `Route`

creators_custom/StopsCreatorFenix.py
```diff
@@ -1,14 +1,84 @@
...
+            i = 0
+            for stop in route.stops:
+                if stop.id in stops:
+                    route.stops[i] = stops[stop.id]
+                else:
+                    stop.get_all_data()
+                    stops[stop.id] = stop
+                    route.stops[i] = stop
+                i += 1
+        elif isinstance(route, RouteMaster):
```
grote: unresolved reference `RouteMaster`

creators_custom/StopsCreatorFenix.py
```diff
@@ -1,14 +1,84 @@
...
+            for sub_route_ref, sub_route in route.routes.iteritems():
+                fill_stops(stops, sub_route)
```
grote: unresolved reference `fill_stops`

xamanu: Fixed

creators_custom/StopsCreatorFenix.py
xamanu: It should be fixed (kind of odd, programming blindly here)

creators_custom/StopsCreatorFenix.py
xamanu: It should be fixed (kind of odd, programming blindly here)

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: It should be fixed (kind of odd, programming blindly here)

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

osm2gtfs.py
```diff
@@ -53,239 +44,52 @@ def main():
...
-    feed_info.feed_start_date = start_date
-    feed_info.feed_end_date = end_date
-    feed_info.feed_version = config['feed_info']['version']
-    schedule.AddFeedInfoObject(feed_info)
+    # Add data to schedule object
+    agency_creator.add_agency_to_schedule(schedule)
+    feed_info_creator.add_feed_info_to_schedule(schedule)
+    routes_creator.add_routes_to_schedule(schedule)
+    #stops_creator.add_stops_to_schedule(schedule)
+    #trips_creator.add_trips_to_schedule(schedule)
```
grote: These should probably be commented in.

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,173 @@
...
+                    if route.ref == DEBUG_ROUTE:
+                        print "END: " + end_time + " at " + str(stop)
+                else:
+                    # timepoint="0" (Times are considered approximate)
+                    trip.AddStopTime(schedule.GetStop(str(stop.id)))
+    #                print "INTER: " + str(stop)
+                i += 1
+
+
+    def interpolate_stop_times(trip):
```
grote: static or self

creators_custom/StopsCreatorFenix.py
grote: Thanks, one more: needs to use self or be static

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,650 @@
...
+        return Stop(stop_id, stop_type)
+
+    def save_routes_to_file(self, outes):
+        routes = creators_custom.ROUTES
+        with open('data/routes.pkl', 'wb') as f:
+            pickle.dump(routes, f, pickle.HIGHEST_PROTOCOL)
+
+    def read_routes_from_file(self):
+        if os.path.isfile('data/route_masters.pkl'):
+            with open('data/routes.pkl', 'rb') as f:
```
grote: I don't know if this has been like this before, but it is a problem anyway:

The path should be both times `data/routes.pkl`.

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,650 @@
...
+            x += cos(lat) * cos(lon)
+            y += cos(lat) * sin(lon)
+            z += sin(lat)
+
+        x = float(x / len(nodes))
+        y = float(y / len(nodes))
+        z = float(z / len(nodes))
+
+        center_lat = degrees(atan2(z, sqrt(x * x + y * y)))
+        center_lon = degrees(atan2(y, x))
```
grote: all these math functions are undefined as well.

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,650 @@
...
+            if isinstance(candidate, overpy.Way):
+                lat, lon = Stop.get_center_of_nodes(candidate.get_nodes(resolve_missing=True))
+                distance = util.ApproximateDistance(
+                    lat,
+                    lon,
+                    self.lat,
+                    self.lon
+                )
+            else:
+                distance = util.ApproximateDistance(
```
grote: unresolved reference to `util`

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,173 @@
...
+        service_sunday.SetWeekendService(False)
+        service_sunday.SetDayOfWeekHasService(6, True)
+
+        # add trips for all routes
+        for route_ref, route in sorted(routes.iteritems()):
+            line = schedule.AddRoute(
+                short_name=route.ref,
+                long_name=route.name,
+                route_type="Bus")
+            line.agency_id = agency.agency_id
```
grote: `agency` is undefined

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,173 @@
...
+            self.add_trips_by_day(schedule, line, service_weekday, route, weekday, WEEKDAY)
+            self.add_trips_by_day(schedule, line, service_saturday, route, saturday, SATURDAY)
+            self.add_trips_by_day(schedule, line, service_sunday, route, sunday, SUNDAY)
+
+    def add_trips_by_day(self, schedule, line, service, route, horarios, day):
+        # check if we even have service
+        if horarios is None or len(horarios) == 0:
+            return
+
+        if isinstance(route, RouteMaster):
```
grote: unresolved `RouteMaster`

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,173 @@
...
+        if route.ref == DEBUG_ROUTE:
+            print "\n\n\n" + str(route)
+            print day + " - " + key
+
+        # get shape id
+        shape_id = str(route.id)
+        try:
+            schedule.GetShape(shape_id)
+        except KeyError:
+            shape = transitfeed.Shape(shape_id)
```
grote: unresolved `transitfeed`
These exist a couple of times, but I guess fixing it once will fix it for all.

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,173 @@
...
+                    print "ADD TRIP " + str(trip.trip_id) + ":"
+                self.add_trip_stops(schedule, trip, route, start_time, end_time, opts)
+
+                # interpolate times, because Navitia can not handle this itself
+                self.interpolate_stop_times(trip)
+
+
+    def add_trip_stops(self, schedule, trip, route, start_time, end_time, opts):
+
+        if isinstance(route, Route):
```
grote: unresolved `Route`

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,650 @@
...
+            stop_type = "way"
+        else:
+            raise RuntimeError("Unknown type: " + str(member))
+
+        return Stop(stop_id, stop_type)
+
+    def save_routes_to_file(self, outes):
+        routes = creators_custom.ROUTES
+        with open('data/routes.pkl', 'wb') as f:
+            pickle.dump(routes, f, pickle.HIGHEST_PROTOCOL)
```
grote: Unfortunately, saving routes does not work.

I don't yet understand what this `creators_custom.ROUTES` should be and how it should work, but it is broken and the argument of this method isn't used.

I'll try to fix that in my local copy, so I can provide you with the pkl files to enable you to work without being blind.

creators_custom/TripsCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: Fixed

creators_custom/TripsCreatorFenix.py
xamanu: Fixed

creators_custom/TripsCreatorFenix.py
xamanu: Fixed

creators_custom/TripsCreatorFenix.py
xamanu: Fixed

creators_custom/RoutesCreatorFenix.py
xamanu: This was my proposed workaround until we discussed #26. Let me try to describe the problem (I hope it's understandable): The current code creates a parallel data structure by using custom classes (RouteMaster, Route, Stop) and those are logically connected, which means that they are potentially created in the RoutesCreator, but used in StopsCreator and TripsCreator. I don't think we can resolve this in this PR, especially if we want to go step by step. Therefore I thought putting it into a global variable (just for now) would be a solution until we go into this part.

xamanu:
I think this is becoming to big to handle properly. Therefore I'm going to close this pull request and will file a new one, which consists in a smaller step to review.

xamanu:
Thanks for the great help here!

Issue #27, opened 2016-10-01
Default implementation for stops creation


xamanu:
The current stops creation is very much custom for Florianópolis. After the abstraction work happening in (#9)  I'd suggest working on a default implementation for stops generation based on OSM's Public Transport Schema 2.

Managua can be a good start as the data quality is very consistent and complying the standard data structure without exceptions.

grote:
It would be nice, if you could make a list where the current implementation deviates from a proper default.

xamanu:
This is very much related to #26, but is a different issue so we can have two smaller pull requests each. But probably the main discussion is going to happen in the other issue. Here it's more about stop specific stuff.

As I said in #26 I haven't started thinking too much about it. And this issue is not to say, that all of Fenix has to be different in the default implementation. We just have to look at it and decide which parts are generic and which aren't.

But for sure the name finding part for stops without a name is very particular to Fenix. The default implementation would probably assume to have bus stop names for all stops in OSM already.

I guess as we move on, more things will be figured out and can be discussed here.

xamanu:
Actually if `TripsCreatorFenix` logic implemented pulling `Stop` objects from `data.stops` based on the stop ids in `Route.stops`, then Fenix could use the currently proposed default `StopsCreator` implementation and get rid of the  functions `add_stops_to_routes()` and `_fill_stops()`

grote:
Good idea!

Issue #26, opened 2016-10-01
Default implementation for route creation


xamanu:
The current route creation is very much custom for Florianópolis. After the abstraction work happening in (#9)  I'd suggest working on a default implementation for route generation based on OSM's Public Transport Schema 2.

Managua can be a good start as the data quality is very consistent and complying the standard data structure without exceptions.

grote:
It would be nice, if you could make a list where the current implementation deviates from a proper default.

xamanu:
This is mainly a placeholder for a task (and a follow up PR to #9). Right now, to keep the #9 PR as small as possible no thinking has been going into this and no default implementation has been defined. This here is the issue to take the taks.

Once we finished with the factory structure, we probably want to do some deeper thinking all together how and what to do in a default implementation and how it differs from Fenix code: What we want to move into a default implementation as it is, what we want to improve in Fenix and move over to the default implementation, and what is so particular that it has to live in Fenix.

Here is a very much brainstormy list I can come up with right now.

- Separate routes, stops and trips creation into easy understandable and encapsulated pieces
- I'd prefer to use transitfeed's (routes, stops and trips objects) instead of using parallel data structure between the different steps of creation
- Not using Portuguese language for variable nor function names
- Not using a custom blacklist of routes to exclude
- Minimize the overpass api calls to respect the sever and avoiding errors like `overpy.exception.OverpassTooManyRequests`
- Not using custom urls related to the regional city

grote:
Thanks for the list!

> Separate routes, stops and trips creation into easy understandable and encapsulated pieces

Definitely!

> I'd prefer to use transitfeed's (routes, stops and trips objects) instead of using parallel data structure between the different steps of creation

The downside of this would be that we would make ourselves totally dependent on their classes which might change without further notice. This bit me in the past in another project.

Also, we have no way to extend/change things if we ever need to. I'd be careful with that approach if it is even possible. But maybe I misunderstand. Let's look at this in detail when the time comes.

> Not using Portuguese language for variable nor function names

Sure, all these should be moved into the FenixCreators.

> Not using a custom blacklist of routes to exclude

I think the `fixme` tag that we discussed elsewhere should already take care of that. 

> Minimize the overpass api calls to respect the sever and avoiding errors like overpy.exception.OverpassTooManyRequests

Something I'd love to have your help with!

> Not using custom urls related to the regional city

[...]

PR #29, opened 2016-10-02
#9 - Factory based abstracted schedule creation

xamanu:
This pull request introduces a creator classes for each part of the GTFS feed (agency, feed_info, routes, stops and trips).

A new `CreatorFactory` allows to overwrite default implementations of each individual creator based on naming convention from the selector property of the config file.

No existing logic has been touched, in order to keep pull requests as small as possible. Discussions about logic should go into the two issues about data structure #30 and refactoring the OsmHelper #31.

There are working default implementations for _agency_ and _feed_info_ because these were already (somehow) generic based on values from the config file.

But there are still no default implementations for routes, stops and trips as their current code exclusively works for the Fenix (Florianópolis) network (issues #26 and #27 will handle this). The default creator classes only define the basic functions that those creator are required to implement.


grote:
Thanks for this! I think this step-size is indeed more manageable to merge. Just too bad we lost so much time yesterday when you were flying blind ;)

Did you see my comment on the old PR about the creator directory structure? Would it be nicer to just have one `creators` folder with a subfolder for each region?

xamanu:
Oh, I missed this comment. I agree I'd prefer to have a subfolder structure! This is much nicer. I will play around with it.

creators_custom/StopsCreatorFenix.py
```diff
@@ -0,0 +1,23 @@
+# coding=utf-8
+
+import sys
+import overpy
+import re
+import pickle
+import transitfeed
```
grote: From here on up, the imports are unused.

creators_custom/RoutesCreatorFenix.py
```diff
@@ -0,0 +1,10 @@
+# coding=utf-8
+
+import sys
```
grote: sys is unused

creators_custom/TripsCreatorFenix.py
```diff
@@ -0,0 +1,204 @@
+# coding=utf-8
+
+import os
```
grote: os is unused

creators_default/StopsCreator.py
```diff
@@ -0,0 +1,17 @@
+# coding=utf-8
+
+import sys
+
+class StopsCreator(object):
+
+    def __init__(self, config):
```
grote: missing `data` object?

creators_default/TripsCreator.py
```diff
@@ -0,0 +1,17 @@
+# coding=utf-8
+
+import sys
+
+class TripsCreator(object):
+
+    def __init__(self, config):
```
grote: missing `data` object?

xamanu: Thanks! Fixed.

creators_default/StopsCreator.py
xamanu: Thanks! Fixed.

creators_custom/TripsCreatorFenix.py
xamanu: Thanks! Fixed.

creators_custom/RoutesCreatorFenix.py
xamanu: Thanks! Fixed.

creators_custom/StopsCreatorFenix.py
xamanu: Thanks! Fixed.

xamanu:
Subfolder structure is implemented in last commit :smile:

grote:
Thanks a lot! Going to merge this now.

If you don't like rebasing hold off any new PR for a moment since I will add a style commit and maybe one other.

xamanu:
Very nice! Thanks for reviewing and merging in.

PR #34, opened 2016-10-06
Allow tags from config file for querying

jamescr:
First little size pull request involved in issue#2.


xamanu:
Love it! But as we are going to use QL anyway and not mix it with other query types (nor APIs) I think the variables could be calles `tags` and `bbox`. What do you think?

jamescr:
I agree with your suggestion. I'll change it.

osmhelper/osm_helper.py
```diff
@@ -19,8 +19,22 @@ class OsmHelper(object):
...
+        # bbox from config file for querying
+        self.bbox = (str(config['query']['bbox']["s"]) + "," +
+                     str(config['query']['bbox']["w"]) + "," +
+                     str(config['query']['bbox']["n"]) + "," +
+                     str(config['query']['bbox']["e"]))
+
+        # tags from config file for querying
+        self.tags = ''
+        for key, value in config["query"].get("tags", {}).iteritems():
+            self.tags += str('["'+key+'" = "'+value+'"]')
```
grote: Just a style nipick: PEP8 asks for spaces around the operators and I agree that it is more readable this way.

osmhelper/osm_helper.py
```diff
@@ -69,25 +84,19 @@ def read_stops_from_file():
...
         api = overpy.Overpass()
-
-        result = api.query("""
-        <query type="relation">
-            <has-kv k="route" v="%s"/>
-            <bbox-query e="%s" n="%s" s="%s" w="%s"/>
-        </query>
-        <print/>
-        """ % (route_type, str(bbox["e"]), str(bbox["n"]), str(bbox["s"]), str(bbox["w"])))
+        query_str = 'relation(%s)%s;(._;);out body;' % (bbox, tags)
```
grote: I can't read and maintain these queries anymore, but I hope for comments when we optimize the queries ;)

How is `<has-kv k="route" v="%s"/>` represented in this new query?

grote: If I run this with the current fenix config file, I get this:
```
  File "./osm2gtfs.py", line 117, in <module>
    main()
  File "./osm2gtfs.py", line 58, in main
    args.refresh_route, "bus", config['query']['bbox'])
  File "/home/user/vcs/osm2gtfs/osmhelper/osm_helper.py", line 306, in refresh_route
    routes = OsmHelper.get_routes(bbox, tags)
  File "/home/user/vcs/osm2gtfs/osmhelper/osm_helper.py", line 99, in get_routes
    result = api.query(query_str)
  File "/home/user/.local/lib/python2.7/site-packages/overpy/__init__.py", line 107, in query
    msgs=msgs
overpy.exception.OverpassBadRequest: Error: line 1: parse error: Unknown query clause 
Error: line 1: parse error: ')' expected - 'bus' found. 
Error: line 1: parse error: An empty query is not allowed 
Error: line 1: parse error: Unknown type "{" 
Error: line 1: parse error: An empty query is not allowed 
Error: line 1: parse error: Unknown type "u"
...
```
Please test your PRs so that the old functionality continues to work. Maybe it would be a good idea to set up a CI for that.

osmhelper/osm_helper.py
jamescr: ok. I forgot to add the spaces there. I'll fixed it.

jamescr:
Although I can not test it now I'm uploading the changes just in case.

osmhelper/osm_helper.py
grote: Now I am getting this.
```
  File "./osm2gtfs.py", line 118, in <module>
    main()
  File "./osm2gtfs.py", line 61, in main
    args.refresh_route, data.bbox, data.tags)
  File "/home/user/vcs/osm2gtfs/osmhelper/osm_helper.py", line 310, in refresh_route
    (bbox, tags, '["ref" = "' + route_ref + '"]'))
TypeError: cannot concatenate 'str' and 'int' objects
```

> The tag "route" is not in the query unless it is added to the config file.

Maybe, the old tags should than be added to the config file. I'm not sure if all routes really have the `public_transport:version` tag. Using the old tags for now, allows me to test this later and remove the tags if everything works.

Please let me know when you had a chance to test all code paths that were touched by this PR.

jamescr: Ok, I'll add to the config file the tag (osm k-v) "route" = "bus".

~~~
  "tags": {
            "route": "bus"
        }
~~~

I'm participating in a Informatics Conference in Chile during this week. I have good Internet access some hours a day but still haven't been able to test all code. Yesterday I tried but I got the Too many requests exception.  

I know is not Ok, but again I'll upload the changes just in case. I'll keep trying to test the code. For sure I could test it next week.

grote:
Thanks! Looks like it is working now.

xamanu:
Cool! This is a great improvement. I'm happy it went in.

Although I found one thing we should do better in the future. The `data = OsmHelper()` line was moved before the refresh arguments block in the `main` function. The constructor of the `OsmHelper` class however queries all routes and stops already. This means, even when updating only the routes (or even one route) or only the stops, everything get's queried anyway. Probably this is not very efficient. And we should move those queries and data tree builders into the respective functions, I guess. Instead of building the whole data tree in the constructor function.

grote:
Ahh that explains the strange behavior I had seen, but thought this to have been introduced earlier already. I'd be happy if somebody provided a PR to fix this.

xamanu:
Yes, it has been introduced before, to create those in the constructor. But the `OsmHelper` object was initiated **after** the refresh logic. I agree this is not very good, because it is not very understandable. I just saw it because I was aware of the problem. I'm happy to provide a PR next week, or if anybody else wants to take it, that's also good :smile:

I created an issue for it: #35

Issue #35, opened 2016-10-12
Make data tree creation efficient


xamanu:
The `data = OsmHelper()` object initiation queries and creates the whole data structure of all routes and stops already in the constructor. 

This is not very efficient, because even when updating only the routes (or even one route) or only the stops, everything get's queried and built anyway.

I guess, we should move those queries and data tree builders into the respective functions, instead of building the whole data tree in the constructor function.

xamanu:
Fixed with latest merge.

Issue #31, opened 2016-10-02
Refactor OsmHelper


xamanu:
The OsmHelper is the part of the script that retrieves data from OpenStreetMap and puts it into the own class/object data structure (to be) discussed in #30. It has included a simple caching functionality that saves data obtained from OpenStreetMap into files and grabs them, if present, to avoid unnecessary hits to the OSM's Overpass API.

I think the current structure of the OsmHelper can be improved and I'm doing a very first draft here, so we have a base for discussion:

**OsmHelper**

```
- config
- routes {route_id, RouteMaster}
- stops {osm_id, Stop}
-------
+ get_routes(): Dic of RouteMaster
+ get_route(String route_id): RouteMaster
+ get_stops(): Dic of Stop
+ get_stop(String osm_id): Stop

- query_routes(): Dic of RouteMaster
- query_route(ref): RouteMaster
- query_stops(): List of Stop
- query_stop(osm_id): Stop

+ cache_refresh(name): void
- cache_write(name, content): void
- cache_read(name): content
```

Some thoughts:
- Using generic read and write caching functions (calling for routemasters, routevariants and stops)
- Use one single cache_refresh function, that takes as an argument if all or a certain part of the cache shall be refreshed.
- All RouteVariants are going to be added to a RouteMaster object, even if there is no master relation in OSM. A simple check creates a RouteMaster either based on the master relation or just on the fly, if non-existant.
- The get-functions are wrappers that check for cached data, return this or query from OSM if data is not cached.

xamanu:
I visualized it in a diagram: 

![osm2gtfs](https://cloud.githubusercontent.com/assets/97706/19020965/fff619b4-88b6-11e6-8ea6-a3f7ce05dbb8.jpg)

grote:
Looks nice!

Where are the methods to refresh a route (and maybe even a stop)? Is this `cache_refresh`? If so, maybe instead of passing a name, have dedicated methods for each type.

I was passing the routes into get_stops() and sometimes chaining these calls, because there's no point getting a stop if it is not part of a route relation, right? Also, you when you refresh you route, this might add new stops that you didn't have before.

The same applies when you refresh a route master, you might get a new route variant and new stops.

xamanu:
Refresh routes, stops and single route would happen through `cache_refesh` with an argument stating what to refresh:

- cache_refresh('routes')
- cache_refresh('route', '255')
- cache_refresh('stops')

[...]

Issue #2, opened 2016-09-17
OSM query improvements


xamanu:
Hi,

I think the current bounding box approach is not ideal. Transit networks can overlap and do not end necessarily at squared boxes.

I'd suggest to run a query to obtain relations with
- type=route
- public_transport:version=2
- network=XY

(and probably maintain the bounding box optional to just make the overpass api query more performant).

Does this make sense? I'd happy to provide a pull request for this. But I also think it's best to get feedback before. Thanks!

grote:
Thanks for creating an issue and discussing first before starting to code! :)

You definitively have a point here.  However, in my area, the `network` tag is not properly maintained and would need to be added to all routes first. Do you have an issue with the current bounding box approach in your area?

Like you say, we should still keep a bounding box for performance reasons. How about allowing to (optionally) also add a `network` tag to the queries?

The `public_transport:version` should also be optional since it is not always well maintained.

xamanu:
Generally, I think we should engage people towards using properly tagging and sticking to standards.

I'm fine with using the `network` tag optionally on queries. This works for the problem just as good as doing it the other way around.

Although the `public_transport:version=2` I suggest to include. Historically, In OSM there are  two data structures for public transport (http://wiki.openstreetmap.org/wiki/Key:public_transport:version) and a lot (!) of inconsistencies. The schema version two is an necessary and collaborative effort to create a standard on how to map public transport om OSM. I don't think it's a problem to include it to existing relations. Especially because JOSM needs it to create and verify public transport data, it should be on the relations, if not they are likely to have a non-standard structure and would likely cause such scripts to fail.

grote:
Alright, do you want to work on adding the optional `network` tag to the queries?

I agree that we should aim at making `public_transport:version=2` mandatory, but I doing this now would make this script stop working for me. So unless there's a way to easily batch-edit all my routes, I think it should be at least possible to exclude this from the query.

Also when people start playing with the script, they might have a similar data quality situation and want to assess its feasibility for their use-case, without editing all their routes manually before they can do that. So being able to at least turn it off is probably a good idea.

xamanu:
Don't worry, with JOSM, it's very easy to add a tag to a bunch of relations at once. 

We could also do a little writeup on how people can assure to have good public transit data in OSM. That would help people to improve data quality easily and we could rely on a common ground of data structure for the script.

xamanu:
[...]

Issue #30, opened 2016-10-02
Data structure


xamanu:
osm2gtfs downloads data from OpenStreetMap (Public Transport Schema version 2) and puts it into transitfeed's structure to generate GTFS.

Currently there is a parallel data structure (mainly living in the classes `RouteMaster`, `Route`, `Stop`) with some regional touch. We already started some discussion about this in #26, but I think it's worth to discuss this separately.

Resuming the past comments on this:

We were considering to use directly transitfeed's (routes, stops and trips objects) instead of a parallel data structure. The downside of this would be that we would make ourselves totally dependent on transitfeed's classes which might change without further notice.

I further think it's not really possible because stops associated to routes (and variants) are connected in transitfeed's objects through the trips. Something osm2gtfs generates at the very end based on stops over routes. So this would just not work, as far as I understand. 

Ergo, we need a parallel structure(!?). So, let's think about this and how this would look like.

xamanu:
Here a first proposal:

**RouteMaster**
```
+ route_id: String
+ route_short_name: String
+ route_long_name: String
+ route_type: String
+ route_desc: String
+ route_color: String
+ route_text_color: String
+ bikes_allowed: String
- routes: List of RouteVariant
-------
+ get_routes(): List of RouteVariant
+ add_route(RouteVariant): void
```

**RouteVariant**
```
+ from: String
+ to: String
+ route_url: String
+ bikes_allowed: String
+ shape: List of Dictionary
- stops: List of String (osm_id of Stop)
-------
+ get_stops(): List of RouteVariant
+ add_stop(Stop): void
+ add_shape(List of Dic): void
```

**Stop**
```
+ osm_id: String
+ stop_id: String
+ name: String
+ lat: String
+ lon: String
```

Some thoughts:

* When querying from OSM we try to obtain as much data possible, but all which is not there (e.g. colors, names of stops, bikes_allowed information, etc.) would be just left blank. It's task of the individual (regional) creators to enhance this information intelligently.
* All RouteVariants are going to be added to a RouteMaster object, even if there is no master relation in OSM. A simple check creates a RouteMaster either based on the master relation or just on the fly, if non-existant.
* Stop objects are not direcly added to the routes, but connected through a list of id's based on osm_ids.

xamanu:
I visualized it in a diagram: 

[...]

PR #36, opened 2016-10-30
OSM query improvements (#2) and OsmHelper refactoring (#31);Fixed data creation (#35)

xamanu:
Hello Thorsten,

I fear you are not going to be very happy.... Please excuse me, despite of our general strategy, this time I was not able to file a _small_ pull request! :cry: I'm really sorry, but I saw no real possibilities in doing the query improvements (#2) without restructuring OsmHelper (#31) or vice-versa. Everything was too much interconnected.

I still made it as small as possible, which means it's focused only on those two issues, and I was not touching data structure and other open issues. Most of the work has been done inside the `OsmHelper` class - that was somehow the magic line I didn't want to cross.

The code now comes with comments and documentation and I hope this makes it more understandable. At least to me, it seems to be much more structured now and a huge improvement. See the `OsmHelper` class:

**Public functions**
- `get_routes()`\- Get routes, no matter if they are already cached or not
- `get_stops()` - Get stops, no matter if they are already cached or not

**Private (internal) functions:**
- `_query_routes()` - Query OSM for routes
- `_query_stops()` - Query OSM for stops
- `_build_route_master()` - Create RouteMaster object
- `_build_route_variant()` - Create Route(Variant) object
- `_build_stop()` - Create Stop object
- `_cache_write()` - One single function that writes object to the file cache
- `_cache_read()` - One single function that reads object from the file cache

Some (minor) things I changed outside of `OsmHelper`:
1. Refactored `add_shape()` in `Route` class, to not query for each shape again and again (part of #2)
2. Removed `get_all_data()` from `Stop`. All data is queried at once and this function is obsolete (part of #2)
3. Moved `from_relation_member()` function from `Stop` class to `_build_stop()` in OsmHelper class  (part of #31)
4. Moved `get_interim_stop_name()` from `Stop` to `StopsCreatorFenix`, because this is custom Fenix code.
5. Temporarily removed the refresh on single route functionality from the `main` wrapper. This was not necessary but it makes review easier for this pull request. I'm happy to open an additional issue for this and re-implement this functionality once this pull request went in. And: refreshing all routes is working faster than one route before, due to the query optimizations... :smile:

In order to get rid of `OsmApi` library in favour of using exclusively `Overpy` we have to use a more recent version of overpy which has not been released officially, yet. But the [important commit](https://github.com/DinoTools/python-overpy/commit/ac7df97ef4302d7eda46a6bb34a887e9b275748c) went in more than one year (!) ago Please make sure that your get the new one. I had to do the following:

```
sudo pip uninstall overpy
sudo pip install -r requirements.txt
```

Make sure you re-cache data by running first:
`python osm2gtfs.py -c fenix.json.example --refresh-all
`
And then:
`python osm2gtfs.py -c fenix.json.example`

For me it runs through for Fenix (yeah! for the first time, since I'm working on this ), but validates that some stops are not used at all. Was this the case before? Maybe you can provide me (privately) with a last run of a resulting GTFS (`br-floripa.zip`), to be able to compare them. This can also be caused by the new queries and the custom blacklisted routes of Fenix code.

It is completely understandable if this bigger pull-request does not go in very quickly and I hope that still everybody is still motivated to test it, give feedback and make it a good progress. Personally I think this is one of the most important steps, as it makes it testable for me (finally). It improves huge parts of data retrieval, caching and readability of the whole process. I hope you also like it. Thanks a lot! Sorry again.


grote:
Wow, Felix! Thanks a lot! After browsing over the changes, this looks really good. I will probably only find time on the weekend to look into this in more detail.

xamanu:
Wonderful, I'm looking forward to hearing your feedback then! Thanks!

xamanu:
I did a couple of changes for better readability of the code:

* Moved caching functionality into a separate `Cache` class.
* Renamed `OsmHelper` to `OsmConnector`.

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+            node(r:"platform")->.nodes;
+            way(r:"platform");
+            node(w);
+
+            /* Select all result sets  */
+            ( .nodes;._; );
+
+            /* Return tags for elements */
+            );out body;""" % (self.tags, self.bbox)
+        return api.query(query_str)
```
grote: This might be the reason why you are getting stops that are not part of any route. There's probably stops that are just not assigned to any route relation within the bounding box.

Doesn't need to happen within this MR, but it would be nice if we could check for those stops before the feed is validated and print useful error messages, so the user has a list of stops to look into.

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+        """
+        # Query stops with platform role from selected relations
+        api = overpy.Overpass()
+        query_str = """(
+            /* Obtain route variants based on tags and bounding box */
+            relation%s(%s);
+
+            /*  Query for related bus stop platforms */
+            node(r:"platform")->.nodes;
+            way(r:"platform");
```
grote: I am afraid not all stops in Floripa are using the platform tag. There's probably still many just using highway=>bus_stop. Those will just not be added with this query? If yes, how can we deal with this? Mass-add the proper tags to all stops that are still missing it?

osmhelper/osm_routes.py
```diff
@@ -56,6 +57,7 @@ def __repr__(self):
         rep += "http://www.consorciofenix.com.br/horarios?q=" + str(self.ref)
         return rep
 
+    # TODO: Move over to Fenix implementation
```
grote: I can do those once we have your MR merged.

osmhelper/osm_routes.py
```diff
@@ -110,49 +112,57 @@ def get_first_alt_stop(self):
...
+                del shape_sorter[0]
+                shape_sorter.reverse()
+                shape_sorter.extend(reversed(way_nodes))
             else:
-                sys.stderr.write("Route has non-matching ways: " + str(self) + "\n")
-                sys.stderr.write("  Problem at: http://www.openstreetmap.org/way/" + str(ways_info[way]["id"]) + "\n")
+                sys.stderr.write(
+                    "Route has non-matching ways: " + str(self) + "\n")
+                sys.stderr.write(
+                    "  Problem at: http://osm.org/way/" + str(way) + "\n")
```
grote: Does this work with https as well? If so, we should probably use that.

osmhelper/osm_routes.py
```diff
@@ -110,49 +112,57 @@ def get_first_alt_stop(self):
...
                 break
 
-        # limit the number of nodes per API request to prevent 414
-        size = 100
-        for nodes_sublist in [nodes[i:i + size] for i in range(0, len(nodes), size)]:
-            nodes_info = osm_api.NodesGet(nodes_sublist)
-            for node in nodes_sublist:
-                self.shape.append({'lat': nodes_info[node]["lat"], 'lon': nodes_info[node]["lon"]})
```
grote: Awesome we can get rid of this! :)

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                     str(config['query']['bbox']["n"]) + "," +
+                     str(config['query']['bbox']["e"]))
+
+        # tags from config file for querying
+        self.tags = ''
+        for key, value in config["query"].get("tags", {}).iteritems():
+            self.tags += str('["' + key + '" = "' + value + '"]')
+        else:
+            # fallback
+            self.tags = '["public_transport:version" = "2"]'
```
xamanu: > print("No tags found for querying from OpenStreetMap.")
> print("Using tag 'public_transport:version=2")

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                        members[rv.tags['ref']
+                                ] = self._build_route_variant(rv, result)
+                    else:
+                        print "Member relation is not a valid route variant:"
+                        print "http://osm.org/relation/" + str(member.ref)
+
+            rm = self._build_route_master(route_master, members)
+            # Make sure ref number is not already taken
+            if rm.ref in self.routes:
+                sys.stderr.write("Ref:" + rm.ref + "already in routes\n")
```
xamanu: What about:

> sys.stderr.write("'Ref' of route_master already taken")
> sys.stderr.write("http://osm.org/relation/" + str(route_master.ref))
> sys.stderr.write("Skipped. Please fix in OpenStreetMap")

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+            else:
+                self.routes[rm.ref] = rm
+
+        # Build routes from variants (missing master relation)
+        for rvid, route_variant in route_variants.iteritems():
+            sys.stderr.write("Routevariant without masters\n")
+            rv = self._build_route_variant(route_variant, result)
+            # Make sure ref number is not already taken
+            if rv.ref in self.routes:
+                sys.stderr.write("Ref:" + rv.ref + "already in routes\n")
```
xamanu: > sys.stderr.write("Route (variant) with existing 'Ref'")
> sys.stderr.write("http://osm.org/relation/" + str(route_variant.ref))
> sys.stderr.write("Skipped. Please fix in OpenStreetMap")

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+            if stop_candidate.role == "platform":
+
+                if isinstance(stop_candidate, overpy.RelationNode):
+                    otype = "node"
+
+                elif isinstance(stop_candidate, overpy.RelationWay):
+                    otype = "way"
+
+                else:
+                    otype = "unknown"
```
xamanu: We use it right below and it's important!

> stops.append(otype + "/" + str(stop_candidate.ref))

It's needed because osm ids are not unique over nodes and ways. So, the id's we use are like `node/329857` and `way/329857`

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+        # Make sure name is not empty
+        if 'name' not in stop.tags:
+            stop.tags['name'] = Stop.NO_NAME
+
+        # Ways don't have coordinates and they have to be calculated
+        if stop_type == "way":
+            (stop.lat, stop.lon) = Stop.get_center_of_nodes(stop.get_nodes())
+
+        s = Stop(stop.id, "node", stop.tags['name'], stop.lat, stop.lon)
+        print s
```
xamanu: I'm happy to take it away. :smile:

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+    def _query_routes(self):
+        """Helper function to query OpenStreetMap routes
+
+        Returns raw data on routes from OpenStreetMap
+
+        """
+        # Query relations of route variants, their masters and geometry
+        api = overpy.Overpass()
+        query_str = """(
+            /* Obtain route variants based on tags and bounding box */
```
xamanu: :+1:

osmhelper/osm_connector.py
xamanu: First, there is a misunderstanding: This is not querying for platform tags, but the platform role in the relation.

osmhelper/osm_connector.py
grote: sounds good!

osmhelper/osm_connector.py
grote: So the problem here is that there are two route master returned from the query with the same ref?

If so, the error message should say that I think. It should list both relations with link to OSM, so the user can quickly find both relations.

osmhelper/osm_connector.py
grote: When you throw an exception, the code below isn't going to be executed.

osmhelper/osm_connector.py
grote: Ahh, still need to learn this new syntax... Thanks for explaining.

osmhelper/osm_connector.py
xamanu: The query is actually looking for relations (the same relations used for the routes based on the tags) and then downloads all objects with the role "platform" of these relations (no matter which tags).

We could think about altering this to download all nodes which are tagged as possible stops. And then try to merge them with the route relations. But what about stops that are not properly tagged, what about stops that are from other networks, not subject of this. I think the common ground is, that we download osm route relations and it's stops and then verify them.

creators/Fenix/StopsCreatorFenix.py
```diff
@@ -1,19 +1,142 @@
...
+        # Get stops information
+        stops = data.get_stops()
 
         # add all stops to GTFS
         for stop in stops.values():
+
+            # If there is no name, query one intelligently from OSM
+            if stop.name == Stop.NO_NAME:
+                self._get_interim_stop_name(stop)
+                print stop
```
xamanu: Yes, actually this is the reason why I moved (recently) the cache into it's own class, so we can easily cache it here. I add a line to the `StopCreatorFenix`

```
if stops_changed:
     # Cache stops with newly created stop names
     Cache.write_data('stops', self.stops)
```

creators/Fenix/StopsCreatorFenix.py
```diff
@@ -1,19 +1,142 @@
...
                 lat=float(stop.lat),
                 lng=float(stop.lon),
                 name=stop.name,
                 stop_id=str(stop.id)
             )
+
+        # Add loose stop objects to route objects
+        self.add_stops_to_routes(schedule, data)
+
+    def add_stops_to_routes(self, schedule, data):
```
xamanu: Yes, no need to have it here.

creators/Fenix/StopsCreatorFenix.py
```diff
@@ -1,19 +1,142 @@
...
+    def _fill_stops(self, stops, route):
+        """Fill a route object with stop objects for of linked stop ids
+
+        """
+        if isinstance(route, Route):
+            i = 0
+            for stop in route.stops:
+                if stop in stops:
+                    # Replace stop id with Stop objects
+                    route.stops[i] = stops[stop]
```
xamanu: Yes, I'm with you. There is a basic new concept here. Let me explain:

Before: A route had it's stops as objects assigned. One identical stop, which has been used by different routes was instantiated several times and was handled as different ones. I didn't like the idea. 

Now: With the query optimizations this is not happening anymore. Routes and Stops are handled separately and efficiently. The route objects usually have a dictionary of (only) stop ids to point to the stop objects.

But Fenix current logic expects them to be objects. I would suggest to change Fenix' logic in the future. This replacement from ids to objects is just a temporary way to simulate the conversion which actually should be improved (in my humble opinion) in Fenix' logic.

creators/Fenix/StopsCreatorFenix.py
```diff
@@ -1,19 +1,142 @@
...
+                i += 1
+
+        elif isinstance(route, RouteMaster):
+            for route_variant_ref, route_variant in route.routes.iteritems():
+                self._fill_stops(stops, route_variant)
+
+        else:
+            raise RuntimeError("Unknown Route: " + str(route))
+
+    def _get_interim_stop_name(self, stop):
```
xamanu: Yes, those are the ones that make it (still) slow. But an optimization of this is clearly out of scope of this PR. This should be handled in a separate issue and though through very well. Then this could be also interesting not only for Fenix but also other networks.

osmhelper/osm_routes.py
xamanu: yeah :smile:

creators/Fenix/StopsCreatorFenix.py
grote: Thanks!

osmhelper/osm_connector.py
xamanu: This is correct. But there are also other roles, or object that should be in there without roles, etc. Personally I think we should encourage people to fix their OpenStreetMap data and not assume all possible failures they are doing.

BTW for poorly maintained routes nobody has schedule information and is using this script to generate GTFS. It's just easier to fix OSM data and better for everybody.

creators/Fenix/StopsCreatorFenix.py
grote: What you call an idea was completely unintentional ;)

I'm fine leaving it like this for now, but please add an TODO that explains what needs to be done.

creators/Fenix/StopsCreatorFenix.py
grote: Yes, the optimization is out of scope, but renaming the method isn't.

osmhelper/osm_connector.py
grote: Also, please don't forget the line breaks when writing to STDERR

osmhelper/osm_connector.py
xamanu: Yes, I'm aware of this. I never expect here to be something else than way nor node. `unknown` would be indeed a serious issue.

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                else:
+                    rv = result.get_relations(member.ref)
+                    if bool(rv):
+                        print "Route variant was assigned again:"
+                        print "http://osm.org/relation/" + str(rv.id)
+                        members[rv.tags['ref']
+                                ] = self._build_route_variant(rv, result)
+                    else:
+                        print "Member relation is not a valid route variant:"
+                        print "http://osm.org/relation/" + str(member.ref)
```
xamanu: Fixed in the updated push.

osmhelper/osm_connector.py
xamanu: Fixed in the updated push.

osmhelper/osm_connector.py
xamanu: Same to the comment above. All this proper error handling would be soo much easier within #30. So, let's not forget it, but also not make it part of the scope of this PR. Before the problem existed the same way, but not even any message was advised.

creators/Fenix/StopsCreatorFenix.py
xamanu: ok. Added in latest update of the commit.

creators/Fenix/StopsCreatorFenix.py
xamanu: Yes, you are right. Fixed in latest update of commit.

osmhelper/osm_routes.py
xamanu: Cool!

creators/Fenix/StopsCreatorFenix.py
```diff
@@ -1,19 +1,150 @@
...
             schedule.AddStop(
                 lat=float(stop.lat),
                 lng=float(stop.lon),
                 name=stop.name,
                 stop_id=str(stop.id)
             )
+
+        if stops_changed:
+            # Cache stops with newly created stop names
+            Cache.write_data('stops', self.stops)
```
grote: I actually meant to cache the stops after every call to `find_best_name_for_unnamed_stop()`. This might seem excessive, but disk access is still way fast than network access. When running your branch, the script terminates quite often when finding the names and with each restart it needs to start from scratch. Would be nice if it could continue where it left off.

grote: Also, what does `self.stops` refer to here?

osmhelper/osm_connector.py
xamanu: Fixed and added line breaks.

creators/Fenix/StopsCreatorFenix.py
xamanu: Yes, good idea to cache every time a name has been found (successfully or not), to be able to continue. This also would allow to get rid of the `stops_changed` flag. I love it. Changed in latest update to the commit.

grote:
I am still running a test for a version from some commits earlier. Has the missing trips issue already been addressed? If not, do you need help finding it?

xamanu:
Thanks for all the good feedback! No, I wasn't able to find and address properly the missing trips issue. Your help on this would be very much appreciated! Hopefully you are more efficient on this, with the Fenix code.

creators/Fenix/StopsCreatorFenix.py
xamanu: Here is the ticket for optimization: #39 (Find best name for unnamed stop)

osmhelper/osm_connector.py
xamanu: Added a [new comment ](https://github.com/grote/osm2gtfs/issues/30#issuecomment-258695414) about the test improvements to #30

osmhelper/osm_connector.py
xamanu: Added a [new comment ](https://github.com/grote/osm2gtfs/issues/30#issuecomment-258695414) about the test improvements to #30

osmhelper/osm_stops.py
```diff
@@ -1,11 +1,7 @@
 # coding=utf-8
 
-import overpy
 import re
-import sys
 from math import cos, sin, atan2, sqrt, radians, degrees
-from transitfeed import util
-
 
 STOP_REGEX = re.compile('(TICAN|TISAN|TICEN|TITRI|TILAG|TIRIO|TISAC).*')
```
xamanu: Hmm, but `normalize_name` is only getting called from `match_first_stops` which only get's called from `TripsCreatorFenix` when everything is already build as it should be. If we start working on this, it would already be related very much to #30... And I wanted to make this PR small :laughing:

grote: Sure, let's keep this small and don't change it any more than really necessary. I am just trying to find the reason, why the trips aren't created. And so far it looks like the stops can't be matched due to the normalization failing. Maybe that is relating to the new way of storing the stop objects?

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,336 @@
...
+                elif isinstance(stop_candidate, overpy.RelationWay):
+                    otype = "way"
+
+                else:
+                    otype = "unknown"
+                    raise RuntimeError("Unknown type: " + str(stop_candidate))
+
+                stops.append(otype + "/" + str(stop_candidate.ref))
+
+        rv = Route(route_variant.id, None, None, stops, None, ref)
```
grote: I think this is the bug. You are just passing None in to and from, but I need those to identify the route variants.

grote: The name though not as critical is also `None` and can probably already be set here.

osmhelper/osm_helper.py
```diff
@@ -1,328 +0,0 @@
...
-            to = None
-
-        # more tags: operator
-
-        print("Adding route %s..." % str(ref))
-
-        if ref in route_masters:
-            # route has a master, so update it in the route_master and add it to the routes
-            stops = route_masters[ref].routes[rid].stops
-            route_masters[ref].routes[rid] = Route(rid, fr, to, stops, route_masters[ref], ref, name)
```
grote: That looks like the critical code you dropped and that is most likely causing the bug.

osmhelper/osm_connector.py
grote: My point is just, you don't need to assign otype anymore right before throwing a runtime exception, because the assignment isn't going to be used anyway ;)

osmhelper/osm_helper.py
xamanu: I made some changes (recent commit update), now the gtfs is way bigger (1.7MB), but probably not as big as the original outcome (2.4MB)

osmhelper/osm_connector.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # Build routes from variants (missing master relation)
+        for rvid, route_variant in route_variants.iteritems():
+            sys.stderr.write("Route (variant) without masters\n")
+            rv = self._build_route_variant(route_variant, result)
+            # Make sure ref number is not already taken
+            if rv.ref in self.routes:
+                sys.stderr.write("Route (variant) with existing 'Ref'\n")
+                sys.stderr.write(
+                    "http://osm.org/relation/" + str(route_variant.ref) + "\n")
```
grote:
```
  File "./osm2gtfs.py", line 111, in <module>
    main()
  File "./osm2gtfs.py", line 58, in main
    data.get_routes(refresh=True)
  File "osm2gtfs/osmhelper/osm_connector.py", line 153, in get_routes
    "http://osm.org/relation/" + str(route_variant.ref) + "\n")
AttributeError: 'Relation' object has no attribute 'ref'
```

grote: `s/route_variant/rv/`?

xamanu: Not sure, but I think it's `s/ref/id/` ? (route_variant.id)

osmhelper/osm_helper.py
xamanu: I get 5501 trips... Have you done `--refresh-routes` or `--refresh-all`?

xamanu: Could this be because oft he missing `RouteMaster` object in the same `Route`(variant) initiation? It's just anoying, because I'm building the `RouteMaster` object after creating the `Route`(variants)...

osmhelper/osm_stops.py
```diff
@@ -33,123 +29,7 @@ def __repr__(self):
...
-            stop_type = "way"
-        else:
-            raise RuntimeError("Unknown type: " + str(member))
-
-        return Stop(stop_id, stop_type)
-
+    # TODO: Move over to Fenix implementation
     @staticmethod
     def normalize_name(old_name):
         name = STOP_REGEX.sub(r'\1', old_name)
```
grote: Could you please add this here as a temporary fix:
```python
name = name.replace('Terminal Centro', 'TICEN')
```
Somebody has renamed this station and broke the normalization https://www.openstreetmap.org/way/43719878#map=18/-27.59847/-48.55306

This gets us some more trips back.

osmhelper/osm_connector.py
grote: crazy indeed. I think you are right and it should be route_variant.id

osmhelper/osm_helper.py
grote: Check out this route: http://www.openstreetmap.org/relation/2079729
I'm getting this for this route:
```
Route (variant) with existing 'Ref'
http://osm.org/relation/2079727
Skipped. Please fix in OpenStreetMap
```
What is the problem with that route?

osmhelper/osm_connector.py
grote: Thanks! I noticed that the fenix config file actually has tags defined, but they are not used. There must be a bug here as well which might also cause some routes from missing that don't have the public_transport:version tag. Let's use the tags from the config file for now and change it later to public_transport:version.

osmhelper/osm_stops.py
xamanu: Added with the latest update on the commit.

osmhelper/osm_connector.py
xamanu: Fixed with the latest commit update.

osmhelper/osm_connector.py
xamanu: Thanks for reporting. There seemed to be an error with the commit before this PR. Should be fixed now with the latest commit update.

osmhelper/osm_helper.py
xamanu: Got it! I was actually replacing `Route`(variants) inside a `RouteMaster` when having the same `ref`. This was not correct and has been fixed in the latest commit update. Thank you for spotting me that! Now I'm at 9092 trips.... That are too much? Was the old code not generating all trips? Or is there still a problem?

osmhelper/osm_connector.py
xamanu: Oh, you are very right :smile: That was not making any sense. Fixed now.

osmhelper/osm_connector.py
xamanu: Fixed.

creators/Fenix/StopsCreatorFenix.py
xamanu: Fixed.

osmhelper/osm_stops.py
xamanu: This seems to be solved by now.

osmhelper/osm_connector.py
xamanu: Done.

osmhelper/osm_helper.py
grote: Wohoo that sounds good! :) Good catch! :+1: 

Let me run that myself and check if there's any outstanding issues.

xamanu:
Ah, I just saw I wrote Thorsten instead of Torsten. Sorry for that!!

osmhelper/osm_helper.py
grote: Did you push that fix? It is still skipping routes like this one http://www.openstreetmap.org/relation/1864112

xamanu: Eh, yes. See line 117 of osm_connector.py
>  members[rv.id] = self._build_route_variant(rv, result)

xamanu: I also have `route_short_name` (`ref`) 843 in routes.txt (`route_id`=69) and 116 trips in trips.txt

grote: Is your HEAD a2be0b3f88cc5a5e861f536199d8de06975792b6 ?

xamanu: Yes. Want to check on [Mumble](http://wiki.openstreetmap.org/wiki/Mumble)?

Issue #39, opened 2016-11-06
Find best name for unnamed stop


xamanu:
Some ideas:

- Improve `find_best_name_for_unnamed_stop()`, if possible
- Move out of Fenix into standard procedure
- Make it configurable to use it or not through config file

xamanu:
Do we want to have it configurable through the config file or as an argument? I'm not sure, but tending to see it nicer as an argument.

grote:
My rule of thumb here is that everything that is likely not going to change over various calls of the script should be in the config file and something that you just want to change for some calls of the script should be a command line option like refreshing the data.

In this case, this rule would probably go for config file as I assume missing names is a problem that either exists in a region or it doesn't.

grote:
However, if it is activated in the config file, there could be a command-line option to turn it off in case you want a quick run-through.

xamanu:
I agree with you thumb rule. But I can think of scenarios that wouldn't work. 

Let's assume stop names are getting found on `--refresh-stops` (as suggested in a recent comment in the latest pull request) and not inside the `StopCreator`:

When finding stop names abort in the process (happens quite often on my side - overpass load error) then there is no way to "continue" from that point.

I prefer a command line argument which runs the stop name finding process independently from refresh stops and stops creation processes and caches as it does at the moment.

xamanu:
Or in config file, maybe even respecting both. But not within refresh-stop logic, it should be independent. That's what I wanted to say.

PR #42, opened 2016-11-11
Improved finding of names for unnamed stops

xamanu:
- Moved finding of names for unnamed stops logic from `FenixStopCreator` into `OsmHelper`
- Setting flag to search for unnamed stops  in config (`name_auto`)
- Introduced name for unnamed stops in config (`name_without`)
- Moved `add_shape()` from `Route` to `_generate_shape()` in `OsmHelper`
- Moved `is_valid_stop_candidate()` from `Stop` to `_is_valid_stop_candidate()` in `OsmHelper`

### DIFF
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -3,6 +3,7 @@
 import sys
 import overpy
 from collections import OrderedDict
+from transitfeed import util
 from core.cache import Cache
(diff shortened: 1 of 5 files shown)


xamanu:
This is related to issue #39

PR #43, opened 2016-11-12
Improved finding of names for unnamed stops - #39

xamanu:
Related to issue #39.

- Moved finding of names for unnamed stops logic from `FenixStopCreator` into `OsmHelper`
- Setting flag to search for unnamed stops  in config (`name_auto`)
- Introduced name for unnamed stops in config (`name_without`)
- Moved `add_shape()` from `Route` to `_generate_shape()` in `OsmHelper`
- Moved `is_valid_stop_candidate()` from `Stop` to `_is_valid_stop_candidate()` in `OsmHelper`

### DIFF
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -3,6 +3,7 @@
 import sys
 import overpy
 from collections import OrderedDict
+from transitfeed import util
 from core.cache import Cache
 from core.osm_routes import Route, RouteMaster
 from core.osm_stops import Stop
@@ -46,6 +47,17 @@ class OsmConnector(object):
             self.tags = '["public_transport:version" = "2"]'
             print("No tags found for querying from OpenStreetMap.")
             print("Using tag 'public_transport:version=2")
+
+        # Define name for stops without one
+        self.stop_no_name = 'No name'
+        if 'stops' in config and 'name_without' in config['stops']:
+            self.stop_no_name = config['stops']['name_without'].encode()
+
+        # Check if auto stop name logic should be used
+        self.auto_stop_names = False
+        if 'stops' in config and 'name_auto' in config['stops']:
+            if config['stops']['name_auto'] == "yes":
+                self.auto_stop_names = True
 
         # Selector
         if 'selector' in config:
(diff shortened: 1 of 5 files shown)


core/osm_connector.py
```diff
@@ -314,7 +338,7 @@ def _build_stop(self, stop, stop_type):
 
         # Make sure name is not empty
         if 'name' not in stop.tags:
-            stop.tags['name'] = Stop.NO_NAME
+            stop.tags['name'] = "[" + self.stop_no_name + "]"
```
grote: Why use the square brackets here? When you don't search for stop names and there's stops without name, there will be square brackets in the name. Is that intentional?

core/osm_connector.py
```diff
@@ -300,8 +324,8 @@ def _build_route_variant(self, route_variant, query_result_set, rm=None):
 
                 stops.append(otype + "/" + str(stop_candidate.ref))
 
-        rv = Route(route_variant.id, fr, to, stops, rm, ref, name)
-        rv.add_shape(route_variant, query_result_set)
+        shape = self._generate_shape(route_variant, query_result_set)
+        rv = Route(route_variant.id, fr, to, stops, rm, ref, name, shape)
```
grote: Nice the shape now gets supplied at construction time!

core/osm_connector.py
```diff
@@ -372,3 +396,170 @@ def _query_stops(self):
...
+        street names and amenities.
+
+        Caches stops with newly guessed names.
+
+        """
+        # Loop through all stops
+        for stop in self.stops.values():
+
+            # If there is no name, query one intelligently from OSM
+            if stop.name == "[" + self.stop_no_name + "]":
```
grote: I see this is how you identify the stops that haven't got a name. Maybe it is better use `None` here and only change all that are still none `None` after this method ran (or didn't ran) to the variable from the config file?

core/osm_connector.py
xamanu: Yes, it is! Because I don't want to hit the nearest search if it was unsuccessful once. This is the same behaviour as before.

When creating the stops from the first query the ones without name get something like `[No name]`, then it tries to guess their name, in case it was not possible to find one the name will be `No name` (without brackets) and next time it would not search (and query overpass api) again.

Issue #14, opened 2016-09-18
Flexibile start and end dates


xamanu:
@grote worte:

> For the dates, we need a more flexible solution, because when used with Navitia, the difference between start and end date can never be more than a year and ideally, for this use-case, I like these dates to automatically move forward when I run the script again without needing to update them somewhere manually.

Great idea. Maybe we can make the script to understand syntax like `now` and `+1year`, etc...

grote:
> Maybe we can make the script to understand syntax like now and +1year, etc..

That might be overkill. How about the following?

```json
"feed_info" : {
  ...
  "validity" : {
    "dates" : "fixed",
    "start_date" : "0000-00-00",
    "end_date" : "0000-00-00"
  }
}
```
and this:
```json
"feed_info" : {
  ...
  "validity" : {
    "dates" : "automatic"
  }
```
Automatic then just takes the current data minus one month as start date and one year later as end date.

*Or* we make it even simpler and use automatic mode automatically when start_date and end_date are not specified. That's probably better.

There should also be a warning if the feed is valid for more than a year.

xamanu:
Good proposal. I think this is the way to go.

xamanu:
Implemented in #45

PR #45, opened 2016-11-12
Flexibile start and end dates - #14

xamanu:
Related to issue #14.

- Moved all config related things from main script into `Configuration` class.
- When no start date is defined in config file it takes the first day of the current month.
- When no end date is defined it uses the start date and builds related to it the end date to the last day of the one month before in one year in the future.
- Validates if start date is not after end date.
- Validates if date range is not more than one year.

### DIFF
--- a/core/configuration.py
+++ b/core/configuration.py
@@ -1,0 +1,145 @@
+# coding=utf-8
+
+import os
+import sys
+import json
+import datetime
+import logging
+from calendar import monthrange
+
+
+class Configuration(object):
+    """The Configuration class validates and prepares configuration data from
+    the config file for further use in the script.
+
+    """
+
+    def __init__(self, args):
+        """Contructor function
+
+        This function gets called when Configuration object are created.
+
+        Based on the configuration from the config file it validates and
+        prepares all mandatory configuration elements.
+
+        """
+        # Load config file from argument of standard location
+        self.config = self._load_config(args)
+
+        # Define name for output file
+        self.output = self._define_output_file(args)
+
+        # Validate and prepare start and end date
+        self._prepare_dates()
+
+    def _load_config(self, args):
+        """Loads the configuration. Either the standard location
+        (config.json) or a location specified as an command argument.
+
+        :return config: A dictionary of configuration data.
+
+        """
+        # Load config json file
+        if args.config is not None:
+            config = Configuration._load_config_file(args.config)
+        elif os.path.isfile('config.json'):
+            with open("config.json") as json_file:
+                config = Configuration._load_config_file(json_file)
+        else:
+            sys.stderr.write("Error: No config.json file found.\n")
+            sys.exit(0)
+
+        return config
+
+    @staticmethod
+    def _load_config_file(configfile):
+        """
+        Loads json from config file
+
+        :return config: A dictionary of configuration data.
+
+        """
+        try:
+            config = json.load(configfile)
+        except ValueError, e:
+            sys.stderr.write('Error: Config json file is invalid.\n')
+            print(e)
+            sys.exit(0)
+
+        return config
+
+    def _define_output_file(self, args):
+        """
+        Defines the filename for GTFS file to write.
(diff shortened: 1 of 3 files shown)


core/configuration.py
```diff
@@ -0,0 +1,131 @@
...
+        print("Using the following end date: " +
+              config['feed_info']['end_date'])
+
+        # Validate dates
+        if start_date > end_date:
+            sys.stderr.write('Error: End dates finishes before start date.\n')
+            sys.exit(0)
+        elif end_date - start_date > datetime.timedelta(days=364):
+            sys.stderr.write(
+                "Dates are describing range of more than one year.\n")
```
grote: Good that this a non-fatal warning as not everybody might want to use the created GTFS with navitia and other consumers might not have this one-year limitation. However, you could write `Warning: ` here like you wrote `Error: `, so it is immediately clear what status this message has.

core/configuration.py
```diff
@@ -0,0 +1,131 @@
...
+        Validate and prepare start and end date.
+        Either from config file or based on current date.
+
+        """
+        config = self.config
+
+        try:
+            start_date = datetime.datetime.strptime(
+                config['feed_info']['start_date'], "%Y%m%d")
+        except:
```
grote: Actually, not working with exceptions here might even be better:
```python

if 'start_date' in config['feed_info']:
  do something with config['feed_info']['start_date']
else:
  make your own date...
```

fenix.json.example
```diff
@@ -22,9 +22,7 @@
     "feed_info": {
         "publisher_name": "Torsten Grote",
         "publisher_url": "https://transportr.grobox.de",
-        "version":  "0.1",
-        "start_date": "20160901",
-        "end_date": "20170831"
```
grote: Nice that this removed here, because I prefer using the dynamic dates. However, I am a bit worried that this feature of manually specifying the dates is now not documented anymore. Do you have an idea where this could be done?

xamanu: We should start documentation somewhere anyway. A lot of things are in code but not documented. Probably the wiki here would be a good place.

core/configuration.py
xamanu: Yes, makes sense. I'll do that.

xamanu:
Yes, I also think this class will open the way to have better verification tests on obligatory configuration parts.

core/configuration.py
xamanu: Exceptions were working very well for me. What's the issue? I don't like the if clauses because we have to check for two things `if 'start_date' in config['feed_info']` and if `datetime.datetime.strptime(config['feed_info']['start_date'], "%Y%m%d")` works successfully.

I'm also very interested if you could describe please what you mean with better style! I like to know how to improve code style! Thank you!

grote: To safe some time I looked if somebody already explained that and there's plenty of articles like this: https://realpython.com/blog/python/the-most-diabolical-python-antipattern/

xamanu: This is hilarious! Thanks for pointing me to that!

> The Most Diabolical Python Antipattern ....
> .... the most self-destructive things a Python developer can write

:smile_cat: I'll fix it.

core/configuration.py
xamanu: Fixed.

core/configuration.py
xamanu: Thanks again! I hope you like it more now.

core/configuration.py
```diff
@@ -0,0 +1,131 @@
...
+        elif os.path.isfile('config.json'):
+            with open("config.json") as json_file:
+                config = self._load_config_file(json_file)
+        else:
+            sys.stderr.write("Error: No config.json file found.\n")
+            sys.exit(0)
+
+        return config
+
+    def _load_config_file(self, configfile):
```
xamanu: ok. Done.

xamanu:
Thanks for reviewing the pull requests and all the great feedback!

Issue #48, opened 2016-12-05
TypeError comparing dates


jamescr:
If the json config file includes in the feed info **only one** of the dates: "start_date" or "end_date" then the other is generated automatically. But when this happen the object type for the generated date is *datetime.date* and the type of the date from config file is *datetime.datetime*. When the dates validation is done an TypeError occurs: «can't compare datetime.datetime to datetime.date»

This could be fixed using *datetime.datetime* in both cases. I'll create a PR is you agree.

grote:
I am not too familiar with that code, but what you describe does sound like a bug to me.

Fenix uses completely automatic dates, so it seems not to be affected by that.

PR #49, opened 2016-12-09
Fix the comparing dates bug (#48)

jamescr:
Changes:
- Fix the bug. 
- prints a info message when config file end date is used


grote:
Thanks @jamescr, looks good to me. @xamanu do you want to have a look as well?

grote:
@xamanu ping!

xamanu:
Yes, I want to have a look at this as well. Thanks. Sorry for taking so long. Will do it today/now.

PR #57, opened 2017-02-12
add incofer provider

jamescr:
Adds the Incofer creators for stops, trips and routes, as well as the config file.

### DIFF
--- a/creators/incofer/__init__.py
+++ b/creators/incofer/__init__.py
@@ -1,0 +1,2 @@
+#!/usr/bin/env python
+# coding=utf-8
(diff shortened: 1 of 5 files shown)


xamanu:
Looks good to me!! (Haven't run the code).

I'm just wondering where the config file should live. If, as proposed here, inside the "creators" folder or on the highest level, like the"fenix" config. Either way I think it should be consistent for both following the same principle.

grote:
We could move the fenix config to the creators folder as well and adapt the example in the README.md, maybe also adding Incofer there.

PR #58, opened 2017-02-25
files update

jamescr:
As commented [here](https://github.com/grote/osm2gtfs/pull/57#issuecomment-280961676), this PR includes:
* move the fenix config file fenix.json.example to creators/fenix/fenix.json
* updates the README file


grote:
Thanks!

I think the script is now generic enough that we don't need a big "**Attention**" anymore in the README, but we can change this later.

Issue #55, opened 2017-02-10
Grouping stops


jamescr:
I'm testing Transportr (a locally compiled version) with incofer data and I notice that some stops are avoided as line change points. Example: instead of changing from line 2 to line 4 in stop A is being done in stop B affecting times of the total trip. I suspect maybe that non grouped stops could be the problem, as stop A of line 2 and stop A of line 4 aren't exactly in the same point but are in the same station:

-84.0693107,Estación Atlántico,9.9349735,4310018677,0
-84.0686079,Estación Atlántico,9.9346699,4628999925,0

Adding support for grouping stops (stations) in the GTFS file seems simple. The GTFS part is solved using the parent_station and location_type of the stops.txt (https://developers.google.com/transit/gtfs/reference/stops-file). And from OSM we could check for the stop_area relation the contains both  stop_position. At the end we should have something like this:

stop_lon,stop_name,stop_lat,stop_id,location_type, parent_station
-84.0693107,Estación Atlántico,9.9349735,4310018677,0, ID
-84.0686079,Estación Atlántico,9.9346699,4628999925,0, ID
lon, Estación Atlántico, lat, ID, 1,

grote:
Strange that it doesn't tell you to walk from stop A2 to stop A4.

But in general, I agree that we should have support for parent_stations and stop_area relations.

Your dataset should be sufficiently small to add the parent station manually and see if that helps with the problem. Maybe it is worth a try.

jamescr:
With the stops grouped in a station Transportr tell me to walk from stop A2 and A4, and the results make more sense. Here are the results:

![photo_2017-02-10_14-42-34](https://cloud.githubusercontent.com/assets/593829/22843554/c45aa866-ef9f-11e6-8291-8aee1c199022.jpg)

![photo_2017-02-10_14-42-43](https://cloud.githubusercontent.com/assets/593829/22843555/c46027e6-ef9f-11e6-8437-1e6de28104ca.jpg)

grote:
Thanks for the screenshots. Looks good. If you want to work on this, maybe let's merge your incofer branch first. It seems to be in a usable state already.

jamescr:
Sure I'll work on this issue.

You mean the incofer branch of public-transport-enabler and Transportr ?

grote:
> Sure I'll work on this issue.

Awesome! :)

> You mean the incofer branch of public-transport-enabler and Transportr ?

No, I meant the incofer provider for osm2gtfs.

[...]

PR #59, opened 2017-03-03
default stops creator

jamescr:
Given that stops creator input data comes from OSM, makes sense to have a default stops creator. By the way, I didn't notice that with the latest changes I did to incofer stops creator, it results to be equal  to the fenix stops creator :wink: .

This will make the code easier to mantain. I prefer to merge this before creating a PR for issue #55.


jamescr:
@xirapo is working on a [stops_creator for the irtramma](https://github.com/xirapo/osm2gtfs/blob/working_on_trips/creators/irtramma/stops_creator_irtramma.py) network, which is exactly as the one of fenix and incofer. This default stops_creator could make the job.

grote:
Good point! We should get this merged eventually. I hope I'll find time soon to make sure this doesn't break the fenix creator.

PR #63, opened 2017-06-28
Fixing : /data directory doesnt exist

xirapo:
Hello ,
im francisco from nicaragua , 
i was trying the script with a config.json from managua when an error show up in the terminal
telling me that the data directory doesnt exist , it work when i create it manually but the script didnt 
check if the directory actually exist , so i know my python is not that advanced but i fix it and now 
the script check if the directory exist if not then the script create it for us and works fine the file are save in the same  directory.

sorry for my bad english.


grote:
Hi @xirapo!

Thanks a lot for your first PR! This is very appreciated!

Your solution includes two lines of the code twice. My suggestion would be to just to this before opening the file:

```python
if not os.path.isdir('data'):
  os.mkdir('data')
```

Also I don't think we need to print about creating this directory.

xirapo:
your welcome @grote  , 
its that correct that was what i did understand from your answer.

creators/managua/managua.json
```diff
@@ -0,0 +1,32 @@
...
+    "feed_info": {
+        "publisher_name": "Inwayswemove.com",
+        "publisher_url": "http://inwayswemove.com",
+        "version":  "0.1",
+        "start_date": "20161101",
+        "end_date": "20171031"
+    },
+    "output_file": "data/ni-managua.zip",
+    "selector": "mapanica"
+}
```
grote: Looks like you included this file by mistance in your PR now.

core/osm_connector.py
```diff
@@ -173,6 +173,8 @@ def get_routes(self, refresh=False):
                 self.routes[rv.ref] = rv
 
         # Cache data
+        if not  os.path.isdir('data'): #checking is the dir exist
+                os.mkdir('data')
         Cache.write_data('routes-' + self.selector, self.routes)
```
grote: I think this code would better be placed in the `Cache.write_data()` method, because this is where it logically belongs to.

style nitpick: two spaces after `not`.

xirapo:
what about know ?? ohh that file was the one i was using for the testing of the script sorry about that

core/osm_connector.py
```diff
@@ -1,5 +1,5 @@
 # coding=utf-8
-
+import os
```
grote: This is not needed anymore here.

creators/managua/managua.json
grote: The file is still part of the MR, it seems.

xirapo:
i just make a mess in my first PR. :)

grote:
No problem. I merged your PR now. Thanks a lot again!

Once you get Managua working, I'd be happy to merge your `creators/managua/managua.json` as well.

cc @xamanu @AltNico

xirapo:
@grote  your welcome and ill keep working on the managua files.

PR #65, opened 2017-08-10
Grouping stops

jamescr:
This PR add support for grouping stops (stations) in the GTFS, as mention in issue  #55.

It works for the incofer stops creators and can be easily extended to other creators (even easier when the Pull request #59 get merged)

### DIFF
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -381,8 +381,7 @@ class OsmConnector(object):
                              " Please fix in OpenStreetMap\n")
             sys.stderr.write("http://osm.org/relation/" +
                              str(relation.id) + "\n")
-            stop_area = StopArea(relation.id, stop_members,
-                                 "Stop area without name")
+            stop_area = StopArea(relation.id, stop_members, self.stop_no_name)
         else:
             stop_area = StopArea(relation.id, stop_members,
                                  relation.tags["name"])
@@ -438,7 +437,7 @@ class OsmConnector(object):
             /* Return tags for elements */
             );out body;
 
-            /* Select stop area relations */   
+            /* Select stop area relations */
             foreach.nodes(
             rel(bn:"platform")["public_transport"="stop_area"];
             out body;
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from transitfeed import util
 from core.cache import Cache
 from core.osm_routes import Route, RouteMaster
-from core.osm_stops import Stop
+from core.osm_stops import Stop, StopArea
 
 
 class OsmConnector(object):
@@ -146,7 +146,8 @@ class OsmConnector(object):
                     else:
                         sys.stderr.write(
                             "Member relation is not a valid route variant:\n")
-                        sys.stderr.write("http://osm.org/relation/" + str(member.ref) + "\n")
+                        sys.stderr.write("http://osm.org/relation/" +
+                                         str(member.ref) + "\n")
 
             rm = self._build_route_master(route_master, members)
 
@@ -178,13 +179,15 @@ class OsmConnector(object):
         return self.routes
 
     def get_stops(self, refresh=False):
-        """The get_stops function returns the data of stops from
+        """The get_stops function returns the data of stops and stop areas from
         OpenStreetMap converted into usable objects.
 
-        Data about stops is getting obtained from OpenStreetMap through the
-        Overpass API, based on the configuration from the config file.
+        Data about stops and stop_areas is getting obtained from OpenStreetMap
+        through the Overpass API, based on the configuration from the config
+        file.
 
-        Then this data gets prepared by building up objects of the class Stops.
+        Then this data gets prepared by building up objects of the class Stops
+        and StopArea (when the Stops are members of a stop_area)
 
         It uses caching to leverage fast performance and spare the Overpass
         API. Special commands are used to refresh cached data.
@@ -193,7 +196,8 @@ class OsmConnector(object):
         :param refresh: A simple boolean indicating a data refresh or use of
             caching if possible.
 
-        :return stops: A dictionary of Stops constituting the obtained data.
+        :return stops: A dictionary of Stops and StopAreas constituting the
+            obtained data.
 
         """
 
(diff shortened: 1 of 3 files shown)


grote:
@jamescr This has now conflicts since the incofer stops creator doesn't exist anymore.

jamescr:
Updated using the default stops creator.

grote:
@nlehuby would you mind reviewing this PR?

creators/stops_creator.py
```diff
@@ -53,12 +77,9 @@ def _fill_stops(self, stops, route):
...
             for stop in route.stops:
-                if stop in stops:
-                    # Replace stop id with Stop objects
-                    # TODO: Remove here and use references in TripsCreatorFenix
-                    route.stops[i] = stops[stop]
-                else:
-                    raise RuntimeError("Unknown stop: " + str(stop))
+                # Replace stop id with Stop objects
+                # TODO: Remove here and use references in TripsCreatorFenix
+                route.stops[i] = self._get_stop(stop, stops)
```
nlehuby: this is a bad pattern, the object type of route.stops is inconsistent:
If you use the default creator, you have a GTFS Stop objects list, if not, you have an OSM2GTFS Stop objects list.
You may want instead to create another object, such as route.gtfs_stops ?

This breaks the Accra creator (that do not use the default creator, and need to get the stops of the routes in the trip creator).

creators/stops_creator.py
```diff
@@ -20,18 +21,41 @@ def add_stops_to_schedule(self, schedule, data):
...
+        self.add_stops_to_routes(data)
 
+    def add_stop(this, schedule, stop, parent_station=None, is_station=False):
             # Add stop to GTFS object
-            schedule.AddStop(
+            stop_obj = schedule.AddStop(
                 lat=float(stop.lat),
                 lng=float(stop.lon),
                 name=stop.name,
                 stop_id=str(stop.id)
```
nlehuby: maybe add an prefix or suffix to the id ?
You could find some edge cases where a relation has the same id as a node : if this happens, you may have issues as the stop_id must be unique.

grote:
@nlehuby Thank you very much for your review! 😃 👍  You found issues I didn't even notice.

creators/stops_creator.py
grote: Good point! Only problem is that it would break (my) existing stop IDs. How about only adding a prefix for stop_areas then?

nlehuby: :+1:

creators/stops_creator.py
jamescr: You are right about the inconsistency but that issue is not part of this PR. The actual "default stops_creator" implementation (already in the master branch) behaves as the fenix and incofer stops_creators, so this inconsistency is inherited from the previous stop creators.

This inconsistency should be a issue to be solve in another pull request.

creators/stops_creator.py
jamescr: Ok. I'll add the prefix "SA" to the stop_area only.

core/osm_connector.py
```diff
@@ -345,6 +364,31 @@ def _build_stop(self, stop, stop_type):
...
+                stop = self.stops.pop("node/" + str(member.ref))
+                stop_members["node/" + str(member.ref)] = stop
+
+        if 'name' not in relation.tags:
+            sys.stderr.write("Stop area without name." +
+                             " Please fix in OpenStreetMap\n")
+            sys.stderr.write("http://osm.org/relation/" +
+                             str(relation.id) + "\n")
+            stop_area = StopArea(relation.id, stop_members,
+                                 "Stop area without name")
```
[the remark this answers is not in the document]
jamescr: OK.

creators/stops_creator.py
jamescr: Indeed its going to break Fenix code as it happened with incofer's. Maybe could be fixed as I did in the incofer trips creator.

creators/stops_creator.py
jamescr: >this is a bad pattern, the object type of route.stops is inconsistent:
>If you use the default creator, you have a GTFS Stop objects list, if not, you have an OSM2GTFS Stop objects list.

After reviewing again the code ( and remembering how it works 🤔  ) I noticed that the inconsistency is not as you said. If you use the default creator you have OSM2GTFS Stop Objects list but if you don't use the default creator you will have a strings list, where each string corresponds to the node ID the corresponds to a stop in OSM.

nlehuby: you are right. Ok for an issue.
Maybe it is what the TODO at the previous was supposed to mean ..

jamescr: Here is the issue #71

creators/stops_creator.py
jamescr: @grote I ran fenix code and it works. I think this is because all `stop_area` relations in [floriapa bbox](http://overpass-turbo.eu/s/s8Y) have only one platform element (all relations have 1 stop and 1 platform as members.). With this data the `stop_creator` will not add stops as *parent stations* (based stop areas). But just in case, I committed a suggested modification to the `trips_creator_fenix` in order to manage eventual *parent stations*.

grote: Thanks for checking the fenix code @jamescr!

> With this data the stop_creator will not add stops as parent stations (based stop areas).

What do you mean by that? The `stop_creator` will just ignore those stop area relations?

In the code, it looks like `len(elem.stop_members) > 1` will still be true in that case.

> But just in case, I committed a suggested modification to the trips_creator_fenix in order to manage eventual parent stations.

This change seems to ignore stop areas completely, right?

Incomplete stop area relations might be common in other data sets as well. Shouldn't the default stop creator handle them well then?

jamescr: >> With this data the stop_creator will not add stops as parent stations (based stop areas).

> What do you mean by that? The stop_creator will just ignore those stop area relations?
> In the code, it looks like len(elem.stop_members) > 1 will still be true in that case.

The overpass query retrieves stops tagged as "platform" and its corresponding "stop_area" relation (with its members). Happens that floriapa "stop_area"' relations has two members, one tagged as "platform" and another tagged as "stop_position". When the `StopArea` object is created, member tagged as `stop_position` is ignored because is not a *valid stop candidate*. Given this, in this case the total `StopArea.stop_members` is 1. 

In this cases, adding the `stop_area` as a "gtfs stop station" and a `stop`as a "gtfs stop" makes no sense, that is why the `stop_area` is ignored and only the  `stop` is added to the stops.txt. If the `stop_area` is updated with more members tagged as `platform`  the creator will include the stop_area and each stop (as is going on incofer case). 


>> But just in case, I committed a suggested modification to the trips_creator_fenix in order to manage eventual parent stations.

> This change seems to ignore stop areas completely, right?

Yes. Was just preventing from a break that will never occur. (I'll explain at the end of the comment).

> Incomplete stop area relations might be common in other data sets as well. Shouldn't the default stop creator handle them well then?

Yes, the stop creator would not break if data is incomplete.

---

The type of the list of `stops` used in both trip_creators is `Stop` never a `StopArea`. That is clear when you look at the `stops_creator._get_stop` method, always will return a `Stop`.

Unfortunately for me I tested incofer trip creator with outdated data, when testing fenix with fresh data and back to incofer also with fresh data, everything was ok with the actual trip creators.

grote: Thanks for the explanation! :)

And now let's get this finally merged!

Issue #12, opened 2016-09-18
Documentation on how to assure data quality of public transport routes in OSM


xamanu:
In order to generate proper GTFS files good data quality in OSM is recommended, if not required.A little writeup/manual would help users how to assure data quality.

I think the wiki here would be the best place to live for this information. Maybe once it's getting to be more complete we could integrate it into learnOSM.org

Just a rough brainstorm list of the contents I suggest to include:

- Use JOSM, the PT_Assistant plugin, and the OSM standard `public_transport:version=2`
- Clean up all verification errors related to PT.
- Make a list of recommended tags to be on the relations and which are used in osm2gtfs

xamanu:
Another thing to include: How to download PT data in JOSM comfortably using this query (or similar):

> [out:xml][timeout:125];
(
 relation["public_transport:version"="2"]["route"="bus"];
 relation["public_transport:version"="2"]["public_transport"="stop_area"];
 node["public_transport"="station"];
 way["public_transport"="station"];
);
(._;>;);
out meta;

grote:
Great idea! Please be invited to start with this in the wiki!

Thanks also for this JOSM query example. Didn't know that was possible.

PR #66, opened 2017-08-16
Accra creator

nlehuby:
* GTFS routes are created with OSM route_master relations
* GTFS Trips are created by frequency, using a `frequency` tag on route_master relations
* OSM route relations are used to enrich GTFS trips (direction_id, headsign)
* Stop areas are created (by name and distance)

TODO : 
- [x] use real data for GTFS frequency opening hours
- [x] use real data for trip travel time

### DIFF
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -290,7 +290,8 @@ class OsmConnector(object):
                 return
 
         name = route_master.tags['name']
-        rm = RouteMaster(route_master.id, ref, name, members)
+        frequency = route_master.tags['frequency']
+        rm = RouteMaster(route_master.id, ref, name, members, frequency)
         print(rm)
         return rm
 
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -327,6 +327,11 @@ class OsmConnector(object):
         else:
             name = None
 
+        if 'travel_time' in route_variant.tags:
+            travel_time = route_variant.tags['travel_time']
+        else:
+            travel_time = None
+
         stops = []
 
         # Add ids for stops of this route variant
@@ -345,7 +350,7 @@ class OsmConnector(object):
                 stops.append(otype + "/" + str(stop_candidate.ref))
 
         shape = self._generate_shape(route_variant, query_result_set)
-        rv = Route(route_variant.id, fr, to, stops, rm, ref, name, shape)
+        rv = Route(route_variant.id, fr, to, stops, rm, ref, name, shape, travel_time)
         print(rv)
         return rv
 
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -350,7 +350,8 @@ class OsmConnector(object):
                 stops.append(otype + "/" + str(stop_candidate.ref))
 
         shape = self._generate_shape(route_variant, query_result_set)
-        rv = Route(route_variant.id, fr, to, stops, rm, ref, name, shape, travel_time)
+        rv = Route(route_variant.id, fr, to, stops,
+                   rm, ref, name, shape, travel_time)
         print(rv)
         return rv
 
@@ -530,7 +531,6 @@ class OsmConnector(object):
         return False
 
     def _get_names_for_unnamed_stops(self):
-
         """Intelligently guess stop names for unnamed stops by sourrounding
         street names and amenities.
 
--- a/core/osm_connector.py
+++ b/core/osm_connector.py
@@ -290,7 +290,9 @@ class OsmConnector(object):
                 return
 
         name = route_master.tags['name']
-        frequency = route_master.tags['frequency']
+        frequency = None
+        if "frequency" in route_master.tags:
+            frequency = route_master.tags['frequency']
         rm = RouteMaster(route_master.id, ref, name, members, frequency)
         print(rm)
         return rm
--- a/core/osm_routes.py
+++ b/core/osm_routes.py
@@ -23,7 +23,7 @@ class BaseRoute(object):
 
 class Route(BaseRoute):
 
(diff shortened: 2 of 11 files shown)


grote:
Wow that was fast! I am currently traveling, but looks like there's still two TODOs open anyway.

Before I can look into this in detail, maybe check out the default stop creator implementation in another PR. Maybe it can be re-used/merged with yours. There's also one for stop areas (grouping stops).
For the trip creator, we maybe can move the interpolation method into the default `TripsCreator` class to reduce code duplication wherever possible.

nlehuby:
Thanks for your feedback.
You can have a look at the generated GTFS, the TODOs are mocks to be removed soon.

About the `TripCreator`, I've modified the default creator, and refactor the 3 creators accordingly.

About the `StopCreator`, we are not sure about the correct way to handle it accordingly to your vision : 
* you can create GTFS stops without any grouping (current default creator, seems a fine default behaviour)
* you can group GTFS stops using OSM StopArea (PR #55 )
* you can group GTFS stops by name and proximity (current PR, Accra creator)
* you can mix it all (using OSM Stop Area and grouping them by name and proximity)
Feel free to have a chat with @prhod in Paris in September about this.

Enjoy your vacations ;)

prhod:
We talked about this PR with @grote, and ended with : 
* A default generation using OSM StopArea only
* A second method that can use both (OSM StopArea and if not available, using distance and/or name)

The name use can be irrelevant in Brazil, because each StopPoint may have unique name (using street names with house numbers).

nlehuby:
As there is no StopArea in Accra I will stick to the name + distance method in this PR.
Maybe you should make an issue so that we have all theses details somewhere when we will refactor the default stop creator ?

grote:
> As there is no StopArea in Accra I will stick to the name + distance method in this PR.

Sure that makes sense!

> Maybe you should make an issue so that we have all theses details somewhere when we will refactor the default stop creator ?

What do you mean exactly?

The idea discussed with @prhod in Paris was to let the default creator use stop area relations, since this is the OSM approach. But then provide another StopCreator that works based on distance. This one doesn't need to live inside the AccraCreator as it is useful for other regions such as Brazil as well. Ideally, it inherits from the other creator or a common super class to reduce code duplication.

prhod:
What @nlehuby is saying is that this PR is about the Accra GTFS generator. This PR shouldn't be blocked by the global enhancement of osm2gtfs with the StopArea default creator we mentioned. 
Would you agree on this steps : 
* creating an issue about the StopArea default creator (I can do it if you're ok with it)
* merging this PR to make this accra generator "official" (with possible enhancements later)
* working on the StopArea default creator afterward, with possible refacto on existing gtfs creators

grote:
I didn't mean to say that this PR is blocked by the stop creator discussion. Do you consider it ready to be reviewed and merged?

> working on the StopArea default creator afterward

Do you mean to say that you want this to be merged before #65 or will it work either way?

nlehuby:
This PR is ready to be reviewed. 

As Accra does not use the stop default creator, the PRs are supposed to be independent, and may be merged in any order.

jamescr:
I agree everything under accra folder is independent, but we should encourage to use as much as possible the default creators.

@nlehuby a couple of doubts:

Do you have a link to OSM wiki or documentation about the default usage of `frequency` and `travel_time` tags? I think the usage is very straightforward but is better to have a documentation at hand (just in case). 

The `frequency` tag should not be a param of route instead of route_master?

nlehuby:
`frequency` and `travel_time` are not standard tags for now, so they do not have a wiki page.
Do you want me to add a link [to the Accra project](https://wiki.openstreetmap.org/wiki/AccraMobile3#Tro_tro_line) ? Or to add explanation about the tags in a comment ? 

The `frequency` could have been on the route just as the `travel_time` tag. But as it is an approximate value of the frequency of the lines, it would have been duplicated on both routes by the contributors.

jamescr:
> frequency and travel_time are not standard tags for now, so they do not have a wiki page.
Do you want me to add a link to the Accra project ? Or to add explanation about the tags in a comment ?

Maybe the wiki is a great place for that info. (see issue #12)

>The frequency could have been on the route just as the travel_time tag. But as it is an approximate value of the frequency of the lines, it would have been duplicated on both routes by the contributors.

I think `frequency` is better in the `routes` (variants) than in the route master. Because the value is not associated to the master, i mean two routes could be related in a route_master but each route with different frequency.

prhod:
For the `frequency` matter, in some cases you are right : the `frequency` is different from a variation to an other. On the other hand, there are some lines (I think for example of long distance `coach`) that don't need variations, and the `frequency` of the forward and backward routes should be the same. Why not setting this property to the `route_master` ?
Speaking of which, a frequency could  also vary on the time of the day ... :-p
Personally, I think both locations are correct, depending on the accuracy expected.

jamescr:
Well, `frequency` on both `route_master` and `route` is OK for me.

grote:
Just to let you know that for osm2gtfs and this MR, I personally don't care how you structure the data within OSM.

grote:
> As Accra does not use the stop default creator, the PRs are supposed to be independent, and may be merged in any order.

`StopsCreatorAccra` is inheriting from the `StopsCreator` though.

> Well, frequency on both route_master and route is OK for me.

What does it mean then if both have a different frequency?

jamescr:
>> Well, frequency on both route_master and route is OK for me.
>What does it mean then if both have a different frequency?

For me, route variant frequency has priority.

grote:
> For me, route variant frequency has priority.

Sounds reasonable, but should be explicitly defined somewhere.

@nlehuby this now has merge conflicts with master.

nlehuby:
rebase ok

PR #70, opened 2017-10-03
Implement data structure

AltNico:
Based on @xamanu's commit I finished the work on osm2gtfs' data structure #30. I'm currently testing with fenix' creator (got banned one time already by Overpass 🙂) and it seems to work. I hope you have good tests testing the creators too. I haven't tested incofer yet but it should work as it has no dependencies to the newly introduced classes.

PR #72, opened 2017-10-04
[WIP] Add default route creator

### LINKED ISSUE #13: Get  Route Information from OSM
Hi there,

the script obtains some attributes from the route from OpenStreetMap, such as reference number (`ref`) and the name. And then there are additional attributes hard-coded in the script, these are:
- route_desc
- route_url
- route_color
- route_text_color

Those should be pulled from OpenStreetMap and not be hard coded.

The [tagging schema](http://wiki.openstreetmap.org/wiki/Relation:route) provides tags on bus routes for `description`, `colour` and we can also add `website`. For the _text color_ we could either come up with a new tag in OSM or define it in the script's config file (#3)

### LINKED ISSUE #26: Default implementation for route creation

AltNico:
The default route creator is based on Incofer's creator and splits up the different properties in small functions which read from the config and the line object by default, but also can be overridden by custom creators easily. This is already done in Incofer's creator but not in Fenix' and Accra's because they do route creation in their trip creators.

Fixes #26.
Fixes #13.


AltNico:
OK, so i have rebased and made tests passing here too, so this is ready to be reviewed. Keep in mind that it is based on #70 and therefore only the last commit is interesting (https://github.com/grote/osm2gtfs/pull/72/commits/b48099701f32d2ec6f15ba968b8458ba5a83afce at the moment).

creators/routes_creator.py
```diff
@@ -25,4 +25,61 @@ def add_routes_to_schedule(self, schedule, data):
...
+        # Get route information
+        lines = data.get_routes()
+
+        # Loop through all lines
+        for line_ref, line in sorted(lines.iteritems()):
+
+            # Add route information
+            route = schedule.AddRoute(
+                route_id=line_ref,
+                route_type=self._get_route_type(line),
```
grote: I am not sure it is a good idea to get this from the config, because there might be providers with different route types like bus, tram and trains. Shouldn't this come from OSM ideally?

creators/incofer/incofer.json
```diff
@@ -12,6 +12,13 @@
           "operator": "Incofer"
         }
     },
+    "routes": {
+        "color": "ff0000",
+        "description": "Esta línea está a prueba",
+        "text_color": "ffffff",
+        "type": "Tram",
+        "url": "http://www.incofer.go.cr/tren-urbano-alajuela-rio-segundo"
+    },
```
grote: This assumes that all lines have the same color, description, etc.
Is this really something for the config? Maybe these should be seen as default values, if no other source of information is available?

grote: I think @xamanu was advocating for storing these things in OSM, if I remember correctly.

creators/routes_creator.py
AltNico: I created different small functions for the different properties to, like you proposed, be able to override them in specific creators later. In these functions, I first only got the information from OSM and overrode the functions in Incofer's creator, but then later thought that it would be cool if one could specify the values in the configuration. But with your other comment, I indeed think that this is overkill and not wanted. I just pushed my old commit which is much more simple.

creators/incofer/incofer.json
jamescr: Indeed He was advocating. (I should add that data to osm)

creators/routes_creator.py
jamescr: I preffer getting the attributes from OSM. In the case of GTFS optional attributes as fallback we can set a default value for all providers and a warning message with help on how to fix it in OSM.  (as an enhacement, when running the script we can set a flag to ignore warnings ).  

For required attributes (I also prefer to get them from OSM and printing the error in case of)  I'll agree to have the fallback value in the config file.

jamescr: Wouldn't be better to handle the attributes in `core.osm_connector.py`?

xamanu:
I suggest closing this PR and file a new pull request based on the recent work (which has been already successfully used to generate GTFS in Nicaragua): https://github.com/mapanica/osm2gtfs/tree/default-creators

AltNico:
@xamanu
> I suggest closing this PR and file a new pull request based on the recent work, which has been already successfully used by us to generate GTFSes in Nicaragua: https://github.com/mapanica/osm2gtfs/tree/default-creators

I agree with you!

PR #77, opened 2017-10-14
Add the ability to ignore a sequence of routes that doesn't have a master linked to

miguelgraz:


core/osm_connector.py
```diff
@@ -293,6 +293,12 @@ def _build_route_master(self, route_master, members):
         frequency = None
         if "frequency" in route_master.tags:
             frequency = route_master.tags['frequency']
+
+        if all(member is None for member in members.values()):
+            sys.stderr.write(
+                "No 'master' could be obtained from members. Skipping.\n")
+            members = {}
+
```
grote: This looks like it is not adding the ability to ignore, but generally ignoring those routes. In principle this is correct, but the OSM data is not everywhere how it should be. The fenix creator for example also works with routes that do not yet have a master. This PR would break it.

xamanu:
I also think we have to be more tolerant of routes not having a route_master but potentially being used for GTFS generation.

Issue #75, opened 2017-10-07
Regression tests


nlehuby:
We need a way to make sure our changes and refactors do not break the current creators.
Maybe a shell script ...

The process could be, for each creator :
* launch osm2gtfs with mocked or persisted OSM input data
* unzip the GTFS output file
* display a diff of all the csv files

grote:
We definitely need those tests. However, I would prefer these to be in python and not shell scripts. It could be a mix of unit and integration tests.

Skippern:
one way of testing could be to have a test folder with text files containing expected Overpass output, and load them on test instead of actually calling Overpass

xamanu:
There has been great progress on this, made by @prhod and @nlehuby. There is a universal structure to include more tests and call them, also in our new CI.

I'm a bit wondering how we will deal with tests that may fail because of changes in the OSM data base. At a first glance, they seem to be problematic, as they might start to fail, even though nothing has changed in code, or the changes don't cause the test to fail. I actually don't have a solution for this, but I think we need to think together to find a good solution.

grote:
Aren't Overpass responses in those great tests mocked, so they are always the same? This should take care of this problem, right?
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

nlehuby:
You are right @grote: the OSM data is mocked so we only test the code and not the data.

xamanu:
Thanks for the clarification. Perfect, that we can mock it up like this.

Regarding tests for creators/cities (like the Accra tests) and tests for core modules, I suggest we create two subdirectories in `tests`:

* `core` - Where tests for the core modules come in.
* `creators` - Where tests for the different creators/cities live.

The naming convention could be most easily: `tests_<SELECTOR>.py` for creators, like for example `tests_gh_accra.py` or `tests_ni_managua`; or in case of core modules it can be `tests_<FILENAME>.py`, like `tests_osm_connector.py` or `tests_osm_configuration.py`.

What do you think?

prhod:
Hello everyone,
@xamanu : I agree with your proposal (2 directories, and the naming convention)
@grote and @nlehuby : to be precise, OSM Data is mocked for stops and routes, but the search for unnamed stops have been disabled because not necessary for Accra. It should be added, maybe as a core function unit test ?

xamanu:
The PR on #83 introduces the testing directory structure for core components as discussed in this thread here.

xamanu:
[...]

PR #79, opened 2017-10-14
Modularize osm2gtfs and add the setup file

nicolasiensen:
As a preparation for writing automated tests, in this commit, we are
creating a root module called `osm2gtfs`.

Having a root module will make it easier to import classes in the test
files.

We added a new setup file in the root folder so we can install osm2gtfs
with pip, since we are now defining the project dependencies there we
removed the requirements file.

Also, after installing this project with pip you can run the script by
calling osm2gtfs in your terminal.

Thanks for @inytar for the help :)

Relates to #75


grote:
Thanks indeed @nicolasiensen for your first MR! It is very much appreciated!

Also thanks @AltNico for the review!

xamanu:
Love this! Thanks a lot, @nicolasiensen.

PR #82, opened 2017-10-24
Provide some metadata in setup.py

xamanu:


setup.py
```diff
@@ -2,6 +2,15 @@
...
 setup(
     name='osm2gtfs',
+    version='0.0.1',
+    description='Turn OpenStreetMap data and schedule information into GTFS',
+    long_description='Use public transport data from OpenStreetMap and external schedule information to create a General Transit Feed (GTFS).',
+    url='https://github.com/grote/osm2gtfs',
+    license='GPLv3',
+    keywords='openstreetmap gtfs schedule public-transportation python',
+    author='Various collaborators: https://github.com/grote/osm2gtfs/graphs/contributors',
+    author_email='t@grobox.de',
```
grote: Do you mind removing this or using another one? I don't like my email address in clear-text in the internet. Too many SPAM-crawlers.

xamanu: Apologies! I didn't want to expose data from you. I took it away. Probably it would be nice to create some mailing list or a mail forwarder. I can offer mapanica.net as a domain for such a forwarder. But I'm also open for any other server/service.

PR #83, opened 2017-10-24
Introduce ISO 3166-2 country codes for creator directories

xamanu:
The more creator implementations we are getting, the more it will be needed to have a bit more consistent naming of those. Therefore I would like to introduce a ISO 3166-2 country code for each creator and focus - if possible - on the city name rather than the network or transport agency.

What do you think?


AltNico:
Having both the country and the town in the name of the creators is great in my opinion. Only problem I see is that there could be different companies in one town which could be solved by appending its name to the existing country-town combination.

xamanu:
Yes, I also had in mind your thought, @AltNico. We could do the following:

**Currently:**

Everybody names the directory of the custom city implementation as they want. Some use the city name, like `accra`, some the network, like `incofer`, and some the operator, like `fenix`.

**This Pull Request**

Introduces a more standardized way: COUNTRYCODE-CITYNAME

However this is not enough, as @AltNico was describing, and also we have the first edge case here: `cr-incofer`, while "incofer" is the network, on purpose I didn't rename it to "San Jose" or anything, so to be not mixed up with bus networks.

**Idea**

I'd be happy to extend this Pull Request to handle multiple hyphens. In this case we can encourage to use: COUNTRYCODE-CITYNAME-NETWORK

This would bring us to:

* br-florianopolis-sim (class name: BrFlorianopolisSim)
* cr-valle-central-incofer (class name: CrValleCentralIncofer)
* ni-managua-irtramma (class name: NiManaguaIrtramma)

The name will become longer but the naming more consistent and it perfectly handles the edge cases. What do you think about this one?

Skippern:
What with Companies serving Multiple cities?

My BR-transcol and BR-seletivo serves Vila Velha, Viana, Cariacica, Vitória and Serra with two lines extending into Guarapari and Fundão, I am also looking at intermunicipal routes with Planeta, covering most of southern Espírito Santo.

I think country code is highly necessary as city or company name might be repeated in different countries, and also for sorting purposes

My suggestion: <ISO country code>[-<city/state/region>][-<company name>]

grote:
I am not actually sure we need special conventions for creator naming at this point.

xamanu:
A separation by country code would be useful. If not this can quickly become long list of confusing names. It's not a big thing, but a nice improvement. I don't understand why this shouldn't go in...?

grote:
I don't say it can't go in. I just think that it is still not a long list of confusing names. Can't we cross that bridge when we get there?

xamanu:
Sure, this is not urgent now. I'm happy to keep it in the pipeline until there are more creators.

As a constraint: it will be more difficult to change the names of the existing directories, the more creators and maintainers are involved. A good and solid practice can save us time later on.

grote:
Indeed, but do we have a consensus on how to standardize the creator names?

Once we have one, I would like to see a test that checks that all existing creators adhere to that standard, so it gets automatically checked when new creators are added.

Also, we have now creator specific tests in `master`. These names should be standardized as well then.

xamanu:
Yes, better to come to an agreement before:

* I would suggest: ` <ISO 3166-2 two letter country code>-<city/state/region>-<network/operator name>` (requiring all three parts)
* All with small letters; only ASCII letters and numbers

grote:
@nlehuby @jamescr what do you think about the name standardization @xamanu is proposing above?

I had some reservations about adding the `<network/operator name>`, because I am planning to merge GTFSs for different operators eventually, but this is probably a post-processing to be done outside the scope of `osm2gtfs`.

ialokim:
I agree that it is necessary to find a naming convention. Also, I think it's an important possibility to add the `<network/operator name>` to distinguish between different operators, but what about leaving this part optional, because in some cases there's no official network or one simple operator name.

In Estelí, the urban buses are served by different operators, but are coordinated by the local government. What would you propose as `network` in this case? `ni-esteli-alcaldia`? 🤔 

I'm also not sure about the `<city/state/region>` part, because we are thinking of adding a GTFS configuration file for the national bus services.

xamanu:
* Yes, I agree, maybe it would be good to leave the network/operator optional: `<ISO 3166-2 two letter country code>-<city/state/region>[-<network/operator name>]`
* If you wanted to unite all of Nicargua into one I would name it `ni-national-mit` (mit = Ministerio de Transporte)

AltNico:
> Yes, I agree, maybe it would be good to leave the network/operator optional: <ISO 3166-2 two letter country code>-<city/state/region>[-<network/operator name>]

👍 for this because I think it is very important to at least the nation code part of the creators name. We should decide this yet, with only few creators, because later it will be harder to do.

nlehuby:
Hi,

`<ISO 3166-2 two letter country code>-<city/state/region>[-<network/operator name>]` seems good, but may not cover all cases.
For instance, I cannot choose a country code for these cases :
* international coaches (if I want to create a Flixbus GTFS from OSM)
* cross border lines (for instance in [Euskadi](http://www.transfermuga.eu/en/pesa/)) 

but I guess `eu-Europe-Flixbus` and `eu-Euskadi-Pesa` would be ok.

We should not overthink this as we don't know yet how OSM2GTFS will be used...so I'm ok with this convention.
For Accra, it would be `gh-accra`, because there is no branded network, and many differents operators.

xamanu:
Seems like we have a consensus on the creator name standardization.

xamanu:
I'm happy to write the requested extension to this PR and the tests, but in order to not go too much in parallel things that are becoming a beast to rebase on each other, I'd prefer to see first those PRs getting in:

* #91 -  Allow schedule source to be specified  
* #89 - New data structure
* TBD - The default trips and routes creators, which will be based on the two PRs mentioned before.
* #92 - Add Esteli, based on the newly introduced standard creators
* #88 - Add Managua, based on the newly introduced standard creators

Thanks.

jamescr:
As @grote I'm planning to incorporate bus routes to the generated GTFS. In this case the name for the creator will be `cr-gam`, GAM stands for Large Metropolitan Area (Gran Area Metropolitana).

xamanu:
Rerolled, now based on PR #99. This now includes also proper tests for naming of files and directories for additional creators.

xamanu:
Rebased on latest upstream master branch (which included #99 and other nice stuff)

xamanu:
This PR is months old and has been rebased several times. I included all remarks (on getting a consensus and including tests for this).

@grote, please let me know when you are considering to merge this. Then I will rebase again. Keeping it up to date without any perspective is a Sisyphus job.

grote:
Sorry @xamanu, this branch has conflicts again :/

xamanu:
Yes, it has. As said in the last comment I will only prepare it again, when it has a clear perspective of being merged. Please confirm, and I'm happy to rebase it to the lastest master.

grote:
Maybe then it is better to get some reviews from the others first since it touches all creators?

xamanu:
@nlehuby, @jamescr and @ialokim may I ask you for a review from your side, please. Thanks!

osm2gtfs/core/creator_factory.py
```diff
@@ -28,83 +28,143 @@ def __repr__(self):
...
+
+        # Define class name
+        if "_" in selector:
+            split_selector = selector.split("_")
+            class_name_selector = str()
+            for part in split_selector:
+                class_name_selector += part.capitalize()
+        else:
+            class_name_selector = selector.capitalize()
+
```
nlehuby: can we avoid to duplicate the class_name_selector creation code in each function ?

osm2gtfs/tests/README.md
```diff
@@ -1,26 +1,48 @@
...
+
+Tests for core components must be named and placed in the following schema:
+  `creators/test_<SELECTOR>.py`
+
+### Accra, Ghana
+
 #### Description
 
-There are 3 unittests for the Accra GTFS generation
+There are 3 unit tests for the Accra GTFS generation
```
nlehuby: actually, this is not an unit test, but an acceptance one.

osm2gtfs/tests/README.md
```diff
@@ -1,26 +1,48 @@
...
 1. Obtain mock-up data on stops from OpenStreetMap and cache it,
 1. Obtain mock-up data on routes from OpenStreetMap and cache it,
 1. Generate GTFS file from using the previously generated cache data.
 
 #### Validation of the GTFS
 
-The generated GTFS is checked against a reference GTFS file (accra_tests.zip.ref)
-in the `tests/fixtures/accra/` folder. For the moment, only the size of each GTFS file is compared to the reference.
+The generated GTFS is checked against a reference GTFS file (gh_accra_tests.zip.ref)
+in the `tests/creators/fixtures_gh_accra/` folder. For the moment, only the size of each GTFS file is compared to the reference.
```
nlehuby: should be `tests/creators/fixtures/gh_accra/` and not `tests/creators/fixtures_gh_accra/`

osm2gtfs/core/creator_factory.py
ialokim: +1 That would be much nicer indeed!

osm2gtfs/creators/ni_esteli/ni_esteli.json
```diff
@@ -30,6 +30,6 @@
         "version":  "0.1"
     },
     "schedule_source": "https://github.com/mapanica/pt-data-esteli/raw/master/timetable.json",
-    "output_file": "data/ni-esteli.zip",
-    "selector": "esteli"
+    "output_file": "data/ni-esteli-gtfs.zip",
+    "selector": "ni_esteli"
```
ialokim: Why did you go for `_` in the selector's case? I think it would be easier and more consistent to use `-` here (as in the output file, for example) too.

osm2gtfs/tests/core/tests_creator_factory.py
```diff
@@ -0,0 +1,92 @@
...
+            for selector in dirs:
+
+                # Check if creator directory fits naming convention
+                self.assertTrue(
+                    self._check_naming_convention(selector),
+                    'The path "creators/' + selector +
+                    ' " doesn\'t fit the naming convention.')
+
+                # Check for existing configuration file
+                config_file_path = path + selector + "/" + selector + ".json"
```
ialokim: Why has it to be `ni_esteli/ni_esteli.json` for the config file? I think it would be a lot more simple to change this naming convention to only `config.json` in all the creators subfolders, so you don't have to type the (potentially long) selector name twice while running _osm2gtfs_.

I see this would be another change added to this PR, but as you're already renaming the json files, I think it would be the right place.

osm2gtfs/tests/core/tests_creator_factory.py
```diff
@@ -0,0 +1,92 @@
...
+                    ") in config file is not equal to the directory name: " +
+                    selector)
+
+                # Loop through available files in creator directories and check
+                # whether they are correctly named
+                for inner_subdir, inner_dirs, inner_files in os.walk(
+                        path + selector):
+
+                    for filename in inner_files:
+                        # Check the pyhton files (and ignore the others)
```
ialokim: *python

osm2gtfs/tests/core/tests_creator_factory.py
```diff
@@ -0,0 +1,92 @@
...
+                                '.py') and not filename == "__init__.py":
+
+                            # Check if selector has been properly applied to
+                            # file name
+                            self.assertTrue(
+                                filename.endswith(selector + '.py'),
+                                "This file is not well named: " + filename)
+
+                            # Check for valid creators
+                            split_filename = filename.split("_")
```
ialokim: This won't work for `feed_info` (in `creators` list above).

osm2gtfs/tests/core/tests_creator_factory.py
```diff
@@ -0,0 +1,92 @@
...
+            # Check for ISO 3166-2 country code
+            if not len(split_selector[0]) == 2:
+                check = False
+
+            # Check for amount of split elements; can only be 2 or three
+            if len(split_selector) < 2 or len(split_selector) > 3:
+                check = False
+        else:
+            return False
+        return check
```
ialokim: Generally, +1 for adding this test!

Correct me if I'm wrong but this only tests the filenames and not the class names as required by the `CreatorsFactory`. Not sure how difficult it would be to implement this tests too.

osm2gtfs/tests/core/tests_creator_factory.py
xamanu: Yes, I agree with you. It would be nicer, but I wanted to keep this PR smallest possible. However as you are also suggesting it, I will apply it to the next rebase here.

xamanu:
Thank you  very much @nlehuby and @ialokim for your reviews. I will incorporate your suggestions soon.

osm2gtfs/creators/ni_esteli/ni_esteli.json
xamanu: Because in python module names underscores are allowed, according to [PEP8](https://www.python.org/dev/peps/pep-0008/), but hyphens aren't. To me it seems much more consistent to use the underscore, because all files and the creator directory only can contain underscores and no hyphens. Anyway the output file can be specified as wished, there is no rule here.

osm2gtfs/tests/core/tests_creator_factory.py
xamanu: Thanks for noticing!

osm2gtfs/tests/core/tests_creator_factory.py
xamanu: Added the desired test.

xamanu:
Here is a new version of this PR for your consideration. It includes the improvements based on the feedback by the reviews of @nlehuby and @ialokim and it is rebased on the latest upstream `master`.

osm2gtfs/creators/ni_esteli/ni_esteli.json
ialokim: OK, I understand, thanks for your explanation!

osm2gtfs/tests/core/tests_creator_factory.py
ialokim: Thanks a lot!

osm2gtfs/tests/core/tests_creator_factory.py
ialokim: Looks a bit ugly, but I'm okay with it.

osm2gtfs/tests/core/tests_creator_factory.py
```diff
@@ -0,0 +1,116 @@
...
+                            try:
+                                module = importlib.import_module(
+                                    ".creators." + selector + "." + filename[:-3],
+                                    package="osm2gtfs")
+                            except ImportError:
+                                correct_class_name_state = False
+
+                            # Snake to camel case to test for correct class names
+                            classname = ''.join(x.capitalize() or
+                                                '_' for x in filename[:-3].split('_'))
```
ialokim: Not sure if this is because of my lack of understanding, but why exactly did you use this `or '_'` part?

xamanu: I took it from [this source](https://www.w3resource.com/python-exercises/re/python-re-exercise-37.php), but it works also with out the `or '_'` part. So, I took it out.

grote: Please don't copy code whose copyright isn't yours into the project.

xamanu:
This seems to be quite ready to be merged in.

@jamescr: This PR renames the creator `incofer` to `cr_gam`, as you mentioned here earlier. Can you please confirm, that you are fine with this renaming?
@nlehuby: Can you have a look again and confirm the PR after applying the suggestions?

Thanks to everybody! :smile:

osm2gtfs/tests/core/tests_creator_factory.py
xamanu: Thanks for the reminder! You are very right!! I was wrongly assuming a public domain permission, which is not the case (only Creative Commons licensed). 

Anyway, with the latest changes it is not coming from any source, but rather a common knowledge based on a web research, the used code passage as it is - adapted to our code - can be found in numerous public places and doesn't belong to one creator.

osm2gtfs/tests/core/tests_creator_factory.py
ialokim: That's okay for me! :+1:

PR #78, opened 2017-10-14
Unit tests coverage for data objects

jcfausto:
Added some unit test coverage to the core/osm_routes.py module. Currently 63% of coverage. 

This is a contribution to the issue #75 done during the #hacktoberfest

PR #85, opened 2017-10-31
Accra integration tests using unittests

prhod:
This PR aim to check Accra generated Data's consistency with osm2gtfs evolutions.
The same tests framework as https://github.com/grote/osm2gtfs/pull/78  is used.
Those tests will ease validations of modifications using Overpass Mocks and an expected GTFS result generated with the provided mocks. 
There are some code duplications, but avoiding them would conflicts with in progress PR ;-)


osm2gtfs/tests/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Accra
+------
+#### Description
+
+There are 3 unittests for the Accra GTFS generation
+
+1. Generation of the cache stops Data from an Overpass result Mock
+* Generation of the cache routes Data from an Overpass result Mock
+* Generation of the GTFS ZIP File using the previously generated cache data
```
grote: The enumeration does seem to be broken here.

xamanu:
Very nice!

grote:
@prhod how about adding these to the CI?

prhod:
Thanks for the merge. I'll create an other PR with the fix of the readme and the addition of the tests in the CI. I'm glad you suggested it, I was'nt confident that it was a good idea ;)

osm2gtfs/tests/README.md
prhod: sorry for that, my markdown editor was ok with this syntaxe

grote:
> I'm glad you suggested it, I was'nt confident that it was a good idea ;)

No, I think it is a good idea. If a new PR breaks Accra, we want to know right away without doing manual tests.

You can create another test stage for unit tests.

xamanu:
Yes, please add to the CI, this gives immediate feedback. Thanks!

prhod:
Readme fix and CI integration available in https://github.com/grote/osm2gtfs/pull/93 :)

Issue #33, opened 2016-10-02
Default implementation for trips creation


xamanu:
The current trips creation is very much custom for Florianópolis. This issue here is the place to discuss how a default implementation for trips creation could and should look like. Probably it's going to be in the direction of @jamescr Incofer trips creation, as he is using a standard json containing schedule information for all stops. Fenix and Managua is doing more fancy and custom stuff to create a schedule, which the can live (partially) in the respective individual creators.

xamanu:
Trips creator default implementation can only happen after #30.

xamanu:
This is part of PR #99.

xamanu:
Resolved with accepted PR #99

Issue #62, opened 2017-03-25
Format for departure/arrival times


Skippern:
My local company [Expresso Lorenzutti](http://www.expressolorenzutti.com.br/horarios.html) have made time tables available in PDF format. I understand it might be difficult to accommodate all forms of publishing time tables, and have therefor started to write a scraper for this. I would like to save it in a simple, common format, that osm2gtfs can read directly. One easy solution would be to save it as JSON. If you have any suggestions to formats I am open.

xamanu:
This is great. Currently we don't have defined a standard format. It is heavily created to the issue #33. And on the next things to do. 

So, I'd be very happy to define together with you a suitable structure.

What is your take on it? The full schedule for each stop as a json object? Or are you thinking about some simplification here? The use cases we applied until now the osm2gtfs script were based on frequencies or starting time on certain stops or stations. And a way to define either the time for the whole itinery or the individual distance between stops.

All approaches can be valid and useful and we probably want to be flexible on this side and support different structures.

Skippern:
The tables I have use starting times on first point/station, circular routes might show a departure time for a mid-point.

I was thinking of creating a single JSON for each start point/route pair, with 3 lists in them, week days, Saturdays, Sundays.

That way osm2gtfs would identify departure from origin in the same way as I understand it harvests from the website in Florianopolis.

xamanu:
Yes, this is the Florianopolis approach and is fine. Lets think aboit how would look like, a suitable structure.

grote:
This sounds indeed like the same information that is available for Florianopolis. There, I am using a [JSON file](http://www.consorciofenix.com.br/api2/linhas.json) from the transport company. However, I do not recommend to copy that JSON structure entirely.

Their basic structure however could work: List of routes/lines. Each route/line is a dictionary of various information such as name, schedule and schedule exceptions.

> I was thinking of creating a single JSON for each start point/route pair, with 3 lists in them, week days, Saturdays, Sundays.

That is already better than the JSON above, but maybe it would be sufficient to to create one json per route/line as there is information that apply to the line/route as a whole like schedule exceptions for holidays or other meta information you might want to include.

Skippern:
A handwritten suggestion based on [039](http://www.expressolorenzutti.com.br/horarios/039.pdf)
```JSON
{
   "ref": "039",
   "origin": "Setiba",
   "destination": "Trevo BR101",
[...]

PR #88, opened 2017-11-28
Add config file for Managua, Nicaragua network

xamanu:
This is the simple config file for the Managua network. As this is going to use the default creators (#26, #27, #33) there is not more needed than merging them in. This PR is similar to #92 and should be merged in after:

* #91 - Allow schedule source to be defined
* #89 - Data structure
* TBD - Standard creators ([almost ready](https://github.com/mapanica/osm2gtfs/tree/default-creators) and waiting for the other two to be approved)

The reason for posting it here, is because this is already working in local setups and should be in the pipeline to get into upstream.

### DIFF
--- a/osm2gtfs/creators/managua/managua.json
+++ b/osm2gtfs/creators/managua/managua.json
@@ -1,0 +1,38 @@
+{
+    "query": {
+        "bbox": {
+            "n": "12.21",
+            "s": "12.04",
+            "e": "-86.12",
+            "w": "-86.38"
+        },
+        "tags": {
+            "route":"bus",
+            "network": "IRTRAMMA",
+            "public_transport:version": "2"
+        }
+    },
+    "stops": {
+        "name_without": "Parada sin nombre - Agregaselo en rutas.mapanica.net",
+        "name_auto": "no"
+    },
+    "agency": {
+        "agency_id": "NI-IRTRAMMA",
+        "agency_name": "Instituto Regulador del Transporte del Municipio de Managua",
+        "agency_url": "http://www.managua.gob.ni/",
+        "agency_timezone": "America/Managua",
+        "agency_lang": "es",
+        "agency_phone": "+505 255-2189",
+        "agency_fare_url": ""
+    },
+    "feed_info": {
+        "publisher_name": "MapaNica.net",
+        "publisher_url": "https://mapanica.net",
+        "version":  "0.1",
+        "start_date": "20171101",
+        "end_date": "20181031"
+    },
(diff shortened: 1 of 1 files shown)


osm2gtfs/creators/managua/managua.json
```diff
@@ -0,0 +1,34 @@
...
+        "agency_lang": "es",
+        "agency_phone": "+505 255-2189",
+        "agency_fare_url": ""
+    },
+    "feed_info": {
+        "publisher_name": "MapaNica.net",
+        "publisher_url": "https://mapanica.net",
+        "version":  "0.1",
+        "start_date": "20171101",
+        "end_date": "20181031"
```
ialokim: `version`, `start_date` and `end_date` should be set by timetable.json, but can stay here too.

osm2gtfs/creators/managua/managua.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "timetable": "https://raw.githubusercontent.com/mapanica/pt-data-managua/master/timetable.json",
+    "output_file": "data/ni-managua-gtfs.zip",
+    "selector": "irtramma"
```
ialokim: Why did you choose an folder named "managua"? We were using "irtramma", and I'm almost certain you should then change the selector too (or isn't it necessary at all while using default creator?)

osm2gtfs/creators/managua/managua.json
xamanu: I'm not sure about this - for me they are more important in the config file, than in the data source. See [comment here](https://github.com/grote/osm2gtfs/issues/62#issuecomment-339120679) and following. The state of the discussion was to allow both.

osm2gtfs/creators/managua/managua.json
xamanu: Hm, I use Managua since quite some time, because more people can understand what Managua is, rather than IRTRAMMA. Generally I would prefer to have a comprehensive list in the folders for the creators, like proposed also in [this pull request](https://github.com/grote/osm2gtfs/pull/83) here.
* Indeed, as you indicate, it is not necessary to change any selector, because Managua is using the default creators.

osm2gtfs/creators/managua/managua.json
ialokim: Okay, perfect. Thanks for the explanation! Leave it here then 😄

xamanu:
Thanks for the review, @ialokim!

grote:
Is this functional before the default creators have been merged? If not, should we hold off with merging it until they have been merged?

xamanu:
Yes, this should wait until the default creators are merged in.

xamanu:
Included into #99. Closing.

PR #89, opened 2017-11-29
Data structure

xamanu:
This is a new pull request for the revised data structure #30 with a lot of wonderful work by @AltNico.
Please help with your review. Thanks! :smile:


grote:
@xamanu thanks for this! Do you know why the CI didn't run for this MR? Do you mind rebasing this on `master`, so we can get the benefit of the Accra integration test to see if it still passes with these changes?

xamanu:
Will rebase and check.

osm2gtfs/core/osm_connector.py
```diff
@@ -259,48 +271,76 @@ def get_stops(self, refresh=False):
...
+                    name=name, route_type=route_type, frequency=frequency,
+                    route_color=colour, route_text_color=text_colour)
+
+        # Add Itinerary objects (route variants) to Line (route master)
+        for itinerary in list(itineraries.values()):
+            try:
+                line.add_itinerary(itinerary)
+            except ValueError:
+                print('Itinerary ID does not match line ID. Please fix in OSM.')
+                print(line.osm_url)
```
grote: These kind of errors should be printed to `stderr` like elsewhere.

osm2gtfs/core/osm_connector.py
```diff
@@ -259,48 +271,76 @@ def get_stops(self, refresh=False):
...
-        rm = RouteMaster(route_master.id, ref, name, members, frequency)
-        print(rm)
-        return rm
-
-    def _build_route_variant(self, route_variant, query_result_set, rm=None):
-        """Helper function to build a RouteVariant object
-
-        Returns a initiated RouteVariant object from raw data
+            
+        colour = "FFFFFF"
```
grote: I think we should be in line with transitfeed's way of writing `color` here (and below).

osm2gtfs/core/osm_connector.py
```diff
@@ -259,48 +271,76 @@ def get_stops(self, refresh=False):
...
+        if "colour" in route_master.tags:
+            colour = OsmConnector.get_hex_code_for_color(route_master.tags['colour'])
```
grote: Do the tags really use `colour` instead of `color`? Is this defined somewhere?

osm2gtfs/core/osm_connector.py
```diff
@@ -240,8 +249,11 @@ def get_stops(self, refresh=False):
             # valid stop_area candidade?
             if 'public_transport' in relation.tags:
                 if relation.tags["public_transport"] == "stop_area":
-                    self.stops["relation/" + str(relation.id)
-                               ] = self._build_stop_area(relation)
+                    try:
+                        self.stops["relation/" + str(relation.id)
+                                   ] = self._build_stop_area(relation)
+                    except RuntimeError:
+                        print('Cannot add stop area', relation.id)
```
grote: We should not catch `RuntimeError`s. Please use a different exception for this. You can even make your own.

osm2gtfs/core/osm_connector.py
```diff
@@ -598,7 +646,7 @@ def _find_best_name_for_unnamed_stop(self, stop):
         winner_distance = sys.maxint
         for candidate in candidates:
             if isinstance(candidate, overpy.Way):
-                lat, lon = Stop.get_center_of_nodes(
+                lat, lon = OsmConnector.get_center_of_nodes(
```
grote: We do you need to prepend method calls with `OsmConnector` when you are in `OsmConnector`?

osm2gtfs/core/stops.py
```diff
@@ -0,0 +1,37 @@
...
+    osm_type = attr.ib()
+    name = attr.ib()
+    lat = attr.ib()
+    lon = attr.ib()
+    gtfs_id = attr.ib(default=osm_id)
+    osm_url = attr.ib(default="http://osm.org/" +
+                      str(osm_type) + "/" + str(osm_id))
+
+
+class StopArea(object):
```
grote: This doesn't need to be annotated with `@attr.s`?

osm2gtfs/core/stops.py
```diff
@@ -0,0 +1,37 @@
+# coding=utf-8
+
+import attr
+
+
+@attr.s
+class Stop(object):
+
+    osm_id = attr.ib()
+    osm_type = attr.ib()
```
grote: Why do we need an `osm_type`? What values can this have?

osm2gtfs/creators/incofer/trips_creator_incofer.py
```diff
@@ -78,6 +79,7 @@ def _get_itinerary_operation(self, itinerary,
                     operations.append("sunday")
         return operations
 
+    # pylint: disable=arguments-differ
     def _create_service_period(self, schedule, operation):
```
grote: Why do the arguments differ here?

osm2gtfs/creators/routes_creator.py
```diff
@@ -13,4 +13,16 @@ def __repr__(self):
...
+        route_type  # Required: From Line
+
+        route_short_name  # Required: To be generated from Line or Itinerary
+        route_long_name  # Required: To be generated from Line or Itinerary
+
+        route_desc # From Line
+        route_url  # From Line
+        route_color # From Line
+        route_text_color  # From Line
+        """
```
grote: What do these comments mean? What do they relate to?

osm2gtfs/core/osm_connector.py
xamanu: Yes, in OpenStreetMap usually British English is used: https://wiki.openstreetmap.org/wiki/Key:colour

osm2gtfs/core/osm_connector.py
xamanu: British... American... now it's time to chose :smile: I also tend to use the American over the British as in most technical environments it is the quasi standard. I would go for `color`, what do you think, @AltNico?

osm2gtfs/core/osm_connector.py
ialokim: What about supporting both? 😄

osm2gtfs/creators/routes_creator.py
xamanu: They were just to not forget important values :smile: Can be taken out.

osm2gtfs/core/stops.py
xamanu: It can be either `node`, `way` or `relation`. Depending on the type different behaviours might be necessary. E.g for a `way` you need to find the center node, for a `relation` you might look into the members.

xamanu:
Thanks a lot for the review, @grote! I just fixed the noted issues.
But I have no idea why CI is not running here.

Please note: When testing, make sure you clean all your cached data.

grote:
Please check your project settings for Integration & Services and make sure Travis CI is listed there and enabled.

osm2gtfs/core/osm_connector.py
```diff
@@ -252,11 +252,11 @@ def get_stops(self, refresh=False):
...
                         self.stops["relation/" + str(relation.id)
                                    ] = self._build_stop_area(relation)
-                    except RuntimeError:
-                        print('Cannot add stop area', relation.id)
+                    except (ValueError, TypeError) as e:
+                        sys.stderr.write('Cannot add stop area: ' + str(relation.id))
 
         # Cache data
-        Cache.write_data('stops-' + self.selector, self.stops)
+        Cache.write_data('stops-' + self.selector, self.routes)
```
grote: routes are written to stops?

osm2gtfs/core/osm_connector.py
```diff
@@ -382,9 +425,15 @@ def _build_stop_area(self, relation):
...
-
+                if "node/" + str(member.ref) in self.stops:
+                    stop = self.stops.pop("node/" + str(member.ref))
+                    stop_members["node/" + str(member.ref)] = stop
+                else:
+                    sys.stderr.write("Stop not found in stops: ")
+                    sys.stderr.write("http://osm.org/node/" +
+                                     str(member.ref) + "\n")
+        if len(stop_members) < 1:
+            raise RuntimeError('Cannot build stop area with no members')
```
grote: Looks like you are still raising a `RuntimeError` here.

xamanu:
Thanks, now it works. Still a bit to fix, but we are close to green :smile: Will comment then.

xamanu:
It seems like the last error raised here by pylint through the Travis CI are part of attr.s and can not be fixed: https://github.com/python-attrs/attrs/issues/215

osm2gtfs/core/stops.py
xamanu: Another reason is also related to #94: `osm_id` is used for different id spaces, depending on `osm_type`. Both together can identify an object in OSM, not only `osm_id`. For the id of the GTFS it is recommended to combine those two in any way.

xamanu:
Closed in favour of this rebased version: #96

Issue #80, opened 2017-10-14
Source for time data


xamanu:
Currently, this repository includes different implementations for cities. Particularly the trip_creators are very much exclusively targeted and programmed for a special data source. However for any new person coming to this script it is almost impossible to find out where to get those data sources, in order to run osm2gtfs with the existing implementations, for example for the sake of testing. As they are so heavily connected, the data (source) on the schedule should be provided, too.

Instead of adding the actual data, I'd suggest to include a `source` field to the configuration file where a valid url to the file containing the time information should be specified. The script would then download the data, save it into the `data` directory and use it. Probably also an argument (similar to the `--refresh-` ones should be implemented, e.g. `--refresh-time-table`).

Skippern:
My repo https://github.com/Skippern/GV-scraper stores the time tables in `times.json`, and can be refreshed with the commands `get_duration.py` and `get_times.py`, that basically will gather the data and build a new `times.json`, if `--refresh-time-table` checks for the presence of `get_duration.py` and `get_times.py` and runs these commands, than I could do a full update in just one command line

xamanu:
As far as I can see we already have [documentation for this one here](https://github.com/grote/osm2gtfs/wiki/Timetable#timetable). But, does it already work?

I think it would be good to have a `timetable` attribute in the configuration file, which specifies the URL to download. And then some minor caching, as described above.

xamanu:
[Working on it](https://github.com/grote/osm2gtfs/compare/master...xamanu:issue-80-timetable)...

AltNico:
You're right, I already have this in the documentation and also a TODO in my trip creator pull request, but like the whole PR it does not work yet.

xamanu:
The great work of @AltNico is in this branch: https://github.com/mapanica/osm2gtfs/tree/timetable
I guess it is almost ready to be filed as a pull request. I'm just wondering about one little thing I wanted to mention here and get your comments on it:

Currently the code checks in the main class, if the standard creators are used - before the script itself knows about it -, then it enforces the timeline to be parsed, valid and loaded. What about checking only if the file exists and having then the timetable (no matter which format) getting tunneled to the trips creator, which is the one in charge to check parsing errors, etc. I'm happy to provide a patch, but didn't want to do it before asking.

grote:
Yes, this looks a bit strange and should be improved. If this is called `data_source` or something similar, then it could be passed into the trips creators and can be `None` by default. So the individual creator can decide what to do with that information.

[...]

PR #91, opened 2017-11-29
Allow schedule source to be specified

xamanu:
Proposed solution for #80.

* Renamed (transitfeed) `schedule` object to `feed` variable name in osm2gtfs
* Introduced a `schedule_source` in the configuration file. Can be either a local path, a url to download any file to be used in the creators to get the time information in, or a "None" (like in Accra) to omit processing of any schedule file.
* Introduced a `--refresh-schedule-source` argument to re-download a cached schedule file.
* Introduced a `schedule_creator`, which allows  the input file to be load, verified and even manipulated - if needed. The simple standard `schedule_creator` takes a regular json file, but others could be supported with an overridden `schedule_creator_*`)
* Added source file information in the configuration file for each of the existing creators
* Make Incofer use the standard `schedule_creator`; make Fenix rely on the `schedule_source` information of the config file.

### DIFF
--- a/osm2gtfs/core/cache.py
+++ b/osm2gtfs/core/cache.py
@@ -18,9 +18,9 @@ class Cache(object):
         hard drive.
 
         """
-        if not os.path.isdir('data'):  # checking is the dir exist
+        if not os.path.isdir('data'):
             os.mkdir('data')
-        with open('data/' + name + '.pkl', 'wb') as f:
+        with open(os.path.join('data', name + '.pkl'), 'wb') as f:
             pickle.dump(content, f, pickle.HIGHEST_PROTOCOL)
 
     @staticmethod
@@ -31,9 +31,40 @@ class Cache(object):
         file on the hard drive.
 
         """
-        if os.path.isfile('data/' + name + '.pkl'):
-            with open('data/' + name + '.pkl', 'rb') as f:
+        filename = os.path.join('data', name + '.pkl')
+        if os.path.isfile(filename):
+            with open(filename, 'rb') as f:
                 content = pickle.load(f)
             return content
         else:
             return {}
+
+    @staticmethod
+    def write_file(name, content):
+        """Function to write cache
+
+        Writes an object (content) with an indicated name to a file on the
+        hard drive.
+
+        """
+        if not os.path.isdir('data'):
+            os.mkdir('data')
+        with open(os.path.join('data', name), 'wb') as f:
+            f.write(content.read())
+
+    @staticmethod
+    def read_file(name):
+        """Function to read cache
+
+        Reads and returns an object (content) with an indicated name from a
+        file on the hard drive.
+
+        :return file: The read file, or an empty dictionary in case no file
+            was found
+        """
+        filename = os.path.join('data', name)
+        if os.path.isfile(filename):
+            with open(filename, 'rb') as f:
+                return f.read()
+        else:
+            return dict()
--- a/osm2gtfs/core/configuration.py
+++ b/osm2gtfs/core/configuration.py
@@ -4,7 +4,9 @@ import os
 import sys
 import json
 import datetime
+from urllib2 import urlopen
 from calendar import monthrange
+from osm2gtfs.core.cache import Cache
 
 
 class Configuration(object):
@@ -23,13 +25,68 @@ class Configuration(object):
 
         """
         # Load config file from argument of standard location
-        self.config = self._load_config(args)
+        self.data = self._load_config(args)
 
         # Define name for output file
         self.output = self._define_output_file(args)
 
         # Validate and prepare start and end date
         self._prepare_dates()
+
+        # Initiate variable for schedule source information
+        self._schedule_source = None
+
+    def get_schedule_source(self, refresh=False):
+        """Loads the schedule source information.
+
+        Loads a schedule source file from either a path or a url specified
+        in the config file
+
+        :return schedule_source: The schedule read from a file.
+
+        """
+
+        if 'schedule_source' not in self.data:
+            return None
+
+        else:
+            source_file = self.data['schedule_source']
+            cached_file = 'schedule-source-' + self.data['selector']
+
+            # Preferably return cached data about schedule
+            if refresh is False:
+                # Check if _schedule_source data is already present
+                if not self._schedule_source:
+                    # If not, try to get _schedule_source from file cache
+                    self._schedule_source = Cache.read_file(cached_file)
+                # Return cached data if found
+                if bool(self._schedule_source):
+                    return self._schedule_source
+
+            # No cached data was found or refresh was forced
+            print("Load schedule source information from " + source_file)
+
+            # Check if local file exists
+            if os.path.isfile(source_file):
+
+                # Open file and add to config object
+                with open(source_file, 'r') as f:
+                    schedule_source = f.read()
+
+            else:
+                # Check if it is a valid url
+                try:
+                    schedule_source_file = urlopen(source_file)
+                except ValueError:
+                    sys.stderr.write(
+                        "Error: Couldn't find schedule_source file.\n")
+                    sys.exit(0)
+                schedule_source = schedule_source_file
+
+        # Cache data
+        Cache.write_file(cached_file, schedule_source)
+        self._schedule_source = schedule_source
+        return self._schedule_source
 
     def _load_config(self, args):
         """Loads the configuration. Either the standard location
@@ -78,8 +135,8 @@ class Configuration(object):
         # Get and check filename for gtfs output
         if args.output is not None:
             output_file = args.output
-        elif 'output_file' in self.config:
-            output_file = self.config['output_file']
+        elif 'output_file' in self.data:
+            output_file = self.data['output_file']
         else:
             sys.stderr.write('Error: No filename for gtfs file specified.\n')
             sys.exit(0)
@@ -92,7 +149,7 @@ class Configuration(object):
         Either from config file or based on current date.
 
         """
-        config = self.config
+        config = self.data
 
         start_date = False
         if 'start_date' in config['feed_info']:
@@ -107,7 +164,7 @@ class Configuration(object):
             # Use first of current month if no start date was specified
             now = datetime.datetime.now()
             start_date = datetime.datetime.strptime(
-                            now.strftime('%Y%m') + "01", "%Y%m%d")
+                now.strftime('%Y%m') + "01", "%Y%m%d")
             config['feed_info']['start_date'] = now.strftime('%Y%m') + "01"
             print("Using the generated start date: " +
                   config['feed_info']['start_date'])
--- a/osm2gtfs/core/creator_factory.py
+++ b/osm2gtfs/core/creator_factory.py
@@ -5,6 +5,7 @@ from osm2gtfs.creators.agency_creator import AgencyCreator
 from osm2gtfs.creators.feed_info_creator import FeedInfoCreator
 from osm2gtfs.creators.routes_creator import RoutesCreator
 from osm2gtfs.creators.stops_creator import StopsCreator
+from osm2gtfs.creators.schedule_creator import ScheduleCreator
 from osm2gtfs.creators.trips_creator import TripsCreator
 
 
@@ -12,8 +13,8 @@ class CreatorFactory(object):
 
     def __init__(self, config):
         self.config = config
-        if 'selector' in config:
-            self.selector = config['selector']
+        if 'selector' in self.config.data:
+            self.selector = self.config.data['selector']
         else:
             self.selector = None
 
@@ -33,10 +34,10 @@ class CreatorFactory(object):
                 package="osm2gtfs")
             agency_creator_override = getattr(
                 module, "AgencyCreator" + selector.capitalize())
-            print "Agency creator: " + selector.capitalize()
+            print("Agency creator: " + selector.capitalize())
             return agency_creator_override(self.config)
         except ImportError:
-            print "Agency creator: Default"
+            print("Agency creator: Default")
             return AgencyCreator(self.config)
 
     def get_feed_info_creator(self):
@@ -47,10 +48,10 @@ class CreatorFactory(object):
                 package="osm2gtfs")
             feed_info_creator_override = getattr(
                 module, "FeedInfoCreator" + selector.capitalize())
-            print "Feed info creator: " + selector.capitalize()
+            print("Feed info creator: " + selector.capitalize())
             return feed_info_creator_override(self.config)
         except ImportError:
-            print "Feed info creator: Default"
+            print("Feed info creator: Default")
             return FeedInfoCreator(self.config)
 
     def get_routes_creator(self):
@@ -61,10 +62,10 @@ class CreatorFactory(object):
                 package="osm2gtfs")
             routes_creator_override = getattr(
                 module, "RoutesCreator" + selector.capitalize())
-            print "Routes creator: " + selector.capitalize()
+            print("Routes creator: " + selector.capitalize())
             return routes_creator_override(self.config)
         except ImportError:
-            print "Routes creator: Default"
+            print("Routes creator: Default")
             return RoutesCreator(self.config)
 
     def get_stops_creator(self):
@@ -75,11 +76,25 @@ class CreatorFactory(object):
                 package="osm2gtfs")
             stops_creator_override = getattr(
                 module, "StopsCreator" + selector.capitalize())
-            print "Stops creator: " + selector.capitalize()
+            print("Stops creator: " + selector.capitalize())
             return stops_creator_override(self.config)
         except ImportError:
-            print "Stops creator: Default"
+            print("Stops creator: Default")
             return StopsCreator(self.config)
+
+    def get_schedule_creator(self):
+        selector = self.selector
+        try:
+            module = importlib.import_module(
+                ".creators." + selector + ".schedule_creator_" + selector,
+                package="osm2gtfs")
+            schedule_creator_override = getattr(
+                module, "ScheduleCreator" + selector.capitalize())
+            print("Schedule creator: " + selector.capitalize())
+            return schedule_creator_override(self.config)
+        except ImportError:
+            print("Schedule creator: Default")
+            return ScheduleCreator(self.config)
 
     def get_trips_creator(self):
         selector = self.selector
@@ -89,8 +104,8 @@ class CreatorFactory(object):
                 package="osm2gtfs")
             trips_creator_override = getattr(
                 module, "TripsCreator" + selector.capitalize())
-            print "Trips creator: " + selector.capitalize()
+            print("Trips creator: " + selector.capitalize())
             return trips_creator_override(self.config)
         except ImportError:
-            print "Trips creator: Default"
+            print("Trips creator: Default")
             return TripsCreator(self.config)
--- a/osm2gtfs/creators/fenix/trips_creator_fenix.py
+++ b/osm2gtfs/creators/fenix/trips_creator_fenix.py
@@ -1,7 +1,6 @@
 # coding=utf-8
 
 import sys
-import json
 import re
 import transitfeed
 from datetime import timedelta, datetime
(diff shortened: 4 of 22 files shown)


xamanu:
This one was approved by the CI :ok_hand:  Should be generally in a good shape (I hope). It completes the osm2gtfs nicely, because now people don't have to look for the creator related schedule source files. They are now automatically getting downloaded and saved locally. Would be happy to see this getting in. Thanks!

nlehuby:
I think the `schedule_source` should be optional (instead of mandatory, with a value of `None` to be ignored). 
This would be more coherent with the other optional parameters (`start_date`).
What do you think ?

xamanu:
For me, the fundamental idea of osm2gtfs is to combine geo-information from OSM with time  information (schedule) to produce a GTFS. Not having a schedule to feed in and to rely on hard coded time information in the trip creator (like currently implemented for Accra) is in my opinion a very edge case, and should not be encouraged by making `schedule_source` optional. However this edge case has been made possible by specifying explicitly `None`.

xamanu:
Some information for potential reviewers: Check the two different commits in this PR. They are reflecting two different changes and are probably easier to review first separately:

* [Rename `schedule` to `feed`, as a result of the conversation in #80](https://github.com/grote/osm2gtfs/pull/91/commits/08dbc64cb31b1f78d4f45440811ea6d518ffec37)
* [Make source configurable and downloadable](https://github.com/grote/osm2gtfs/pull/91/commits/7789b4c6620c2be6936ff2d557d4c88cb1029be3)

osm2gtfs/core/cache.py
```diff
@@ -37,3 +37,30 @@ def read_data(name):
...
+
+    @staticmethod
+    def write_file(name, content):
+        """Function to write cache
+
+        Writes an object (content) with an indicated name to a file on the
+        hard drive.
+
+        """
+        if not os.path.isdir('data'):  # checking is the dir exist
```
grote: there's a typo in the comment, but the method name would be clear enough for my taste.

osm2gtfs/core/cache.py
```diff
@@ -37,3 +37,30 @@ def read_data(name):
...
+            f.write(content.read())
+
+    @staticmethod
+    def read_file(name):
+        """Function to read cache
+
+        Reads and returns an object (content) with an indicated name from a
+        file on the hard drive.
+
+        """
```
grote: Maybe you should add that it returns an empty `dict` if the file wasn't found?

osm2gtfs/core/cache.py
```diff
@@ -37,3 +37,30 @@ def read_data(name):
...
+    def write_file(name, content):
+        """Function to write cache
+
+        Writes an object (content) with an indicated name to a file on the
+        hard drive.
+
+        """
+        if not os.path.isdir('data'):  # checking is the dir exist
+            os.mkdir('data')
+        with open('data/' + name, 'wb') as f:
```
grote: While that works on Linux, it is better to use `os.path.join()` here.

osm2gtfs/core/cache.py
```diff
@@ -37,3 +37,30 @@ def read_data(name):
...
+    @staticmethod
+    def read_file(name):
+        """Function to read cache
+
+        Reads and returns an object (content) with an indicated name from a
+        file on the hard drive.
+
+        """
+        if os.path.isfile('data/' + name):
+            with open('data/' + name, 'rb') as f:
```
grote: While that works on Linux, it is better to use `os.path.join()` here. You can also introduce a variable for the path, since you use it twice.

osm2gtfs/core/configuration.py
```diff
@@ -31,6 +33,64 @@ def __init__(self, args):
...
+        Loads a schedule source file from either a path or a url specified
+        in the config file
+
+        :return schedule_source_file: The schedule source file object.
+
+        """
+
+        if 'schedule_source' not in self.config:
+            sys.stderr.write("Error: No schedule source specified in config file.\n")
+            sys.exit(0)
```
grote: Like @nlehuby I prefer this to return `None` instead of bailing out.

osm2gtfs/core/configuration.py
```diff
@@ -31,6 +33,64 @@ def __init__(self, args):
...
+            # No cached data was found or refresh was forced
+            print("Load schedule source information from " +
+                  self.config['schedule_source'])
+
+            # Check if local file exists
+            if os.path.isfile(self.config['schedule_source']):
+
+                # Open file and add to config object
+                with open(self.config['schedule_source'], 'r') as schedule_source_file:
+                    self._schedule_source_file = schedule_source_file
```
grote: Does this leak the file object outside the `with` environment and leave it open?

osm2gtfs/core/creator_factory.py
```diff
@@ -5,15 +5,16 @@
...
 
 
 class CreatorFactory(object):
 
     def __init__(self, config):
         self.config = config
-        if 'selector' in config:
-            self.selector = config['selector']
+        if 'selector' in self.config.config:
+            self.selector = self.config.config['selector']
```
grote: config.config looks confusing. Why does this happen now?

osm2gtfs/creators/schedule_creator.py
```diff
@@ -0,0 +1,53 @@
...
+import sys
+import json
+
+
+class ScheduleCreator(object):
+    """
+    The ScheduleCreator loads, validates and prepares configuration data from
+    the schedule source file for further use in the script.
+
+    """
```
grote: Maybe you could at the URL to the documentation here?

osm2gtfs/core/creator_factory.py
xamanu: Because the `config` object of `Configuration` contains an attribute `config` with the processed information from the configuration file. In this PR I started giving the full `config` object to the creators, because some of them might want to use functions attached to `Configuration`, for example the newly introduced `schedule_creator` does this, to load the schedule information when needed.

grote: If you want to do this, you might want to rethink the naming at the same time, so people reading the code have a better change at knowing what is what.

osm2gtfs/core/configuration.py
xamanu: I'm still strongly in favour of explicitly opting out. This goes down to the very basic vision of the script, e.g. look at osm2gtfs' subtitle *Turn OpenStreetMap data **and schedule information** into GTFS*. 

To have an implementation that accepts by default configuration files without any schedule source for me is not a good practice, kind of against the idea of GTFS and encourages to produce low quality GTFS. I do think it should be possible to do so, but then you have to explicitly state this in the configuration file.

osm2gtfs/core/creator_factory.py
xamanu: What about `config.data`? Or is this too generic?

osm2gtfs/core/configuration.py
grote: But weren't advocating for putting as much information into OSM? For frequency based regions, I don't see why they shouldn't define the frequency of a line in OSM and don't have to bother to specify this source.

If you allow `None` here, you might as well allow it to not be there.

osm2gtfs/core/creator_factory.py
grote: I guess `data` is fine, don't have a better idea right now. You forgot to adapt the `param` docstrings where the Configuration object is passed, btw.

osm2gtfs/core/configuration.py
xamanu: Yes, as much information that should be in OSM. And especially to stick to the standards by the community. And full schedule information is clearly not desired to be in OpenStreetMap. For frequency there is no accepted nor proposed tag (standard), that I'm aware of. In other words, this is not expected and mostly not wanted to be added to OSM (unfortunately). Exactly because of this rationale I'm advocating to provide a standard in osm2gtfs to get it in at the moment when it is the right place to feed it in.

Maybe we can agree on a compromise:

* We make `schedule_source` optional.
* The standard `schedule_creator` (not the configuration class) throws an error, if there is no file specified.

This would basically follow your opinion, but (at least for the standard creator) do the same checks, as it does currently.

osm2gtfs/core/configuration.py
xamanu: Yes, is there any better way to do it?

xamanu:
Thanks for the review, @grote ❤️

osm2gtfs/core/configuration.py
grote: Sounds like a good compromise to me. @nlehuby?

osm2gtfs/core/configuration.py
grote: I am not sure it is a good idea to pass file objects around this way. Why not passing its content around instead?

xamanu: Because, we don't know at this point the format of the schedule file. It can be json, it can be csv or any other format, which then will be used in the (overridden) `schedule_creator`

grote: Good point. In this case, maybe just use a string and leave it to the creator to interpret it. So it can handle file opening and closing itself.

xamanu:
Because of other changes, this discussion and question got collapsed. So, I'm posting it here again. It is about the question about the issue risen by @nlehuby in one of the first comments - whether `schedule_source` should be optional.

I proposed a compromise:
>* We make schedule_source optional.
>* The standard schedule_creator (not the configuration class) throws an error, if there is no file specified.

> This would basically follow your opinion, but (at least for the standard creator) do the same checks, as it does currently.

@grote answered and asked:
> Sounds like a good compromise to me. @nlehuby?

osm2gtfs/core/cache.py
xamanu: Yes, thanks, done in the whole class.

osm2gtfs/core/cache.py
xamanu: ok.

osm2gtfs/creators/schedule_creator.py
xamanu: Added

osm2gtfs/core/configuration.py
xamanu: I went for reading the files already at this point, but not decoding it. This is necessary to do proper caching and it avoids passing around any file objects. Hopefully this is also fine?!

grote: Yes, thanks!

xamanu:
This pull request is now at the stage to be tested again - all feedback has been incorporated. The `schedule_source` is currently implemented as proposed: It is optional, but the standard `schedule_creator` checks for a respective json file. For Accra the `source_creator` has been overridden to not process any file.

osm2gtfs/core/creator_factory.py
```diff
@@ -81,6 +82,20 @@ def get_stops_creator(self):
...
+        try:
+            module = importlib.import_module(
+                ".creators." + selector + ".schedule_creator_" + selector,
+                package="osm2gtfs")
+            schedule_creator_override = getattr(
+                module, "ScheduleCreator" + selector.capitalize())
+            print "Schedule creator: " + selector.capitalize()
+            return schedule_creator_override(self.config)
+        except ImportError:
+            print "Schedule creator: Default"
```
grote: There's two python2 prints here.

osm2gtfs/core/configuration.py
```diff
@@ -23,14 +25,69 @@ def __init__(self, args):
...
+        # Initiate variable for schedule source information
+        self._schedule_source = None
+
+    def get_schedule_source(self, refresh=False):
+        """Loads the schedule source information.
+
+        Loads a schedule source file from either a path or a url specified
+        in the config file
+
+        :return schedule_source_file: The schedule source file object.
```
grote: still talking about file object.

osm2gtfs/core/cache.py
```diff
@@ -31,9 +31,40 @@ def read_data(name):
...
+            f.write(content.read())
+
+    @staticmethod
+    def read_file(name):
+        """Function to read cache
+
+        Reads and returns an object (content) with an indicated name from a
+        file on the hard drive.
+
+        :return file: The opened file object, or an empty dictionary in case no
```
grote: still talking about file object.

osm2gtfs/core/creator_factory.py
xamanu: Not only two, all of the prints in this class. Changed now.

osm2gtfs/creators/fenix/trips_creator_fenix.py
```diff
@@ -54,15 +54,16 @@ def __init__(self, config):
...
-        schedule.AddServicePeriodObject(self.service_sunday)
+        feed.AddServicePeriodObject(self.service_weekday)
+        feed.AddServicePeriodObject(self.service_saturday)
+        feed.AddServicePeriodObject(self.service_sunday)
 
         # Get Fenix data from JSON file
         json_data = []
-        with open('data/linhas.json') as f:
+        schedule_source = self.config.get_schedule_source()
+        with open(schedule_source) as f:
```
grote: Did you test this?

xamanu: No, this part I was not able to test. Fenix doesn't run until then, same problem as on master. So, I reverted the changes on this part. Loads the standard file from location.

xamanu:
I tested with Accra, but you are right, I didn't run the tests, yet. There were some minor issues, that I fixed now.

osm2gtfs/core/configuration.py
xamanu: Oh yes, taken out.

osm2gtfs/core/cache.py
xamanu: Oh yes, taken out.

osm2gtfs/creators/fenix/trips_creator_fenix.py
xamanu: I just applied a fix for Fenix. At least it runs now.

xamanu: It works as before, so this is hopefully not a blocker to merge this in.

xamanu: ok, I finally had the chance to have enough stops cached to get until this point in code execution in Fenix. And I guess we can remove the whole block and replace it with `linhas = data.schedule['data']`. This change is now in the PR.

grote:
When running this branch, I still get:
```
  File "/home/user/.local/bin/osm2gtfs", line 11, in <module>
    load_entry_point('osm2gtfs', 'console_scripts', 'osm2gtfs')()
  File "/home/user/vcs/osm2gtfs/osm2gtfs/osm2gtfs.py", line 72, in main
    stops_creator.add_stops_to_feed(feed, data)
  File "/home/user/vcs/osm2gtfs/osm2gtfs/creators/stops_creator.py", line 35, in add_stops_to_feed
    self.add_stop(feed, elem)
  File "/home/user/vcs/osm2gtfs/osm2gtfs/creators/stops_creator.py", line 49, in add_stop
    stop_dict["stop_id"] = str(stop.osm_id)
AttributeError: 'Stop' object has no attribute 'osm_id'
```
Also, there's an unused import lying around still.

grote:
A tip for testing with fenix: If you hit overpass API, disable the stop name search. This is mostly not necessary for testing.

xamanu:
Oh, yes, `osm_id` is already from #96, fixed and tested with Fenix.

grote:
I can confirm that Fenix works now (although the feed looks way smaller than list time I tried it, need to look into that some day). The Accra tests also pass.

However, Incofer from @jamescr crashes. I think before merging in more PRs, we should first add more tests. Otherwise, we are going to screw things up I'm afraid.

xamanu:
> I think before merging in more PRs, we should first add more tests.
Not nice, to draw this line now :cry:

grote:
> (Incofer fails on master as it does on this one)

Oh, I didn't know that. I indeed get the same error message on `master`.

So for me this PR is good to go in.

> Not nice, to draw this line now 😢

I am not drawing a line, but merely suggest that we should maybe focus our energies on adding more tests first before merging even more complex PRs. This one already had some issue we only spotted with code review and manual testing. The others will most likely be even more disruptive.

Not all creator maintainers have always time available to review PRs on short notice to make sure their creators are unaffected. Having more tests would reduce the burden on everybody, because we could rely on the CI to tell us whether something broke or not.

xamanu:
Yes, I also think, the better and more test coverage we have the better we can maintain this growing script! Probably we should consider to require creators with respective tests, as Accra wonderfully did, from all network implementers.

I suggest to create a separate issue to discuss this with all implementers(?)/maintainers(?)/sponsors(?) of the network creators.

grote:
We have #75 for that. I just increased its priority.

I am just leaving this one open for a bit longer in case one of the others wants to review as well.

osm2gtfs/core/configuration.py
```diff
@@ -23,14 +25,69 @@ def __init__(self, args):
...
+        """
+
+        if 'schedule_source' not in self.data:
+            return None
+
+        else:
+            source_file = self.data['schedule_source']
+            cached_file = 'schedule-source-' + self.data['selector']
+
+            # Preferably return cached data about routes
```
ialokim: typo: shouldn't it say `cached data about schedule`?

osm2gtfs/creators/schedule_creator.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, config):
+        self.config = config
+
+    def __repr__(self):
+        rep = ""
+        if self.config is not None:
+            rep += str(self.config) + " | "
+        return rep
+
+    def add_schedule_to_data(self, schedule, data):
```
ialokim: The schedule argument will be overridden in the next line and is never passed (see `osm2gtfs.py`).

osm2gtfs/osm2gtfs.py
```diff
@@ -44,58 +46,62 @@ def main():
     elif args.refresh_stops:
         data.get_stops(refresh=True)
         sys.exit(0)
+    elif args.refresh_schedule_source:
+        config.get_schedule_source(refresh=True)
+        sys.exit(0)
```
ialokim: Why do you `sys.exit(0)` here? Couldn't the script continue to run as in the case of `args.refresh_all`?

osm2gtfs/core/configuration.py
xamanu: Yes, thanks for noting.

osm2gtfs/creators/schedule_creator.py
xamanu: oh, yes, you are right. fixed.

osm2gtfs/osm2gtfs.py
xamanu: Hmm, this is a good question. I follow your point, removed all `sys.exit(0)` from the other calls also.

xamanu:
Thanks for the review @ialokim!

xamanu:
Rebased on latest pull request #93.

xamanu:
All green - all good. Any chance to get it in, anytime soon? 

As a side note: Not having this inside osm2gtfs causes currently significant extra work to keep this in-sync with the other enhancements we are currently working on to get Managua ready. This is mainly because of the renaming request from #80.

grote:
For me it is fine, just waiting a bit with merging to give others a chance to review.

You should be able to rebase other work on top of this branch in the meantime.

xamanu:
> You should be able to rebase other work on top of this branch in the meantime.

Yes, but in our case, [it hurts](http://www.jarrodspillers.com/git/2009/08/19/git-merge-vs-git-rebase-avoiding-rebase-hell.html) (see at the very bottom):
> Don’t rebase branches you have shared with another developer.
> Do otherwise at your own peril - yes, it will hurt.

xamanu:
> You should be able to rebase other work on top of this branch in the meantime.

Yes, but what I'm trying to express is that [it hurts](http://www.jarrodspillers.com/git/2009/08/19/git-merge-vs-git-rebase-avoiding-rebase-hell.html) (last sentences of the article):

> Don’t rebase branches you have shared with another developer.
> Do otherwise at your own peril - yes, it will hurt.

grote:
The same kind of advise applies to force pushing. There's not really a difference between force pushing to this branch and rebasing another branch on this one. Both is easier when you are the only person working on these branches though.

grote:
Thanks for the review @prhod!

xamanu:
Wuhuu! Thanks all!

PR #92, opened 2017-11-29
Add config file for Esteli, Nicaragua network

xamanu:
This PR is provided by @AltNico. It is in a similar situation like #88 and should be merged in after:

* #91 - Allow schedule source to be defined
* #89 - Data structure
* TBD - Standard creators ([almost ready](https://github.com/mapanica/osm2gtfs/tree/default-creators) and waiting for the other two to be approved)

The reason for posting it here, is because this is already working in local setups and should be in the pipeline to get into upstream.


xamanu:
Included into #99. Closing.

PR #93, opened 2017-12-01
accra tests - fix readme and add tests to CI

prhod:


.travis.yml
```diff
@@ -15,3 +15,6 @@ jobs:
     - stage: pylint
       before_script: pip install pylint
       script: pylint --disable=C,R osm2gtfs
+
+    - stage: accra non regression test
+      script: python osm2gtfs/tests/accra_unittests.py
```
grote: Hmm.. could you make this generic, so it is named and functions in a way that it runs *all* unit tests? also those from the future like #78?

prhod: Sorry for answering so late. I'll try to make this generic asap :)

grote: I've seen people use tox, but this is probably the easiest for now: https://setuptools.readthedocs.io/en/latest/setuptools.html#test

I tried `python setup.py test --test-suite=osm2gtfs.tests` and it worked. So just adding the `test_suite` to `setup.py` should do the trick.

prhod:
Hi all, I managed to generalize the execution of tests. As I have an specific order for tests of the Accra test suite, I had trouble making it work with a pylint compatibility.
I ended with a hack in the tests_accra.py load_tests function. Tell me if you have a hint on the proper way of doing it... or if you think it's sufficient for the moment :)

xamanu:
I'm wondering what are the variables `loader`, `tests` and `pattern` are used for?

osm2gtfs/tests/tests_accra.py
```diff
@@ -158,9 +158,14 @@ def test_gtfs_from_cache(self):
...
 
 
-def suite():
-    tests = ['test_refresh_routes_cache', 'test_refresh_stops_cache', 'test_gtfs_from_cache']
-    return unittest.TestSuite(map(TestAccra, tests))
+def load_tests(loader, tests, pattern):
+    # pylint require to use the variables
+    loader = loader
+    tests = tests
+    pattern = pattern
```
grote: You should be able to do something like `# pylint: disable=unused-varialbes` here instead.

osm2gtfs/tests/README.md
```diff
@@ -8,8 +8,8 @@ Accra
 There are 3 unittests for the Accra GTFS generation
 
 1. Generation of the cache stops Data from an Overpass result Mock
-* Generation of the cache routes Data from an Overpass result Mock
-* Generation of the GTFS ZIP File using the previously generated cache data
+1. Generation of the cache routes Data from an Overpass result Mock
```
xamanu: I kindly suggest to rephrase a bit for better language and usage of upper case words. My proposal (feel free to take, adapt or use anything else):

> * Obtain mock-up data on stops from OpenStreetMap and cache it.
> * Obtain mock-up data on routes from OpenStreetMap and cache it.
> * Generate GTFS file from using the previously generated cache data. (remove the "and" from the last list item).

osm2gtfs/tests/tests_accra.py
```diff
@@ -158,9 +158,14 @@ def test_gtfs_from_cache(self):
...
+    test_cases = ['test_refresh_routes_cache', 'test_refresh_stops_cache', 'test_gtfs_from_cache']
```
xamanu: To me this line seems to be very close to the individual needs of Accra, and I guess other networks and other tests regarding some core components might need to have another logic for the test cases. I would like to see them living in the individual test suite.

I'm not a unit test expert in Python! And maybe I'm on a wrong track, question: wouldn't it better/easier to have one "defined and initial" test case for all kind of types of tests (we could call it `test_summary` or anything else), basically a wrapper around all the test cases this is implementing. This will allow other test suits, that are very different (e.g. testing core components will be different than testing a creator set), to implement their test cases, also in a certain order, etc...

osm2gtfs/tests/tests_accra.py
xamanu: I'm wondering what are the variables `loader`, `tests` and `pattern` used for? Do we need them? (sorry might be an ignorant question)

grote: I think they are just used here to make pylint happy which can also be achieved with something similar to what I proposed above.

osm2gtfs/tests/tests_accra.py
xamanu: Sorry, this is already in `tests_accra.py`. So never mind this comment.

osm2gtfs/tests/tests_accra.py
xamanu: Ah, they are required by the discovery call. Thanks.

prhod: @xamanu : variables are described here (and are used by the discovery) : https://docs.python.org/2/library/unittest.html#load-tests-protocol
@grote : thanks for the proposal, I'll check it out asap :)

xamanu:
Looks good to me. Thanks for this great enhancement!

grote:
Thanks also from my side, @prhod! :)

prhod:
thanks to both of you ! Next step : https://github.com/grote/osm2gtfs/pull/78 :)

Issue #76, opened 2017-10-07
Unify travel_time and duration


jamescr:
What is the difference between `travel_time` and `duration` attributes of `Route` object ?

(Sorry for the delayed question related with  PR #66.)

nlehuby:
Here is my guess 
* travel_time is read from an OSM tag whereas duration is set from another source.
* travel_time is an number of minutes whereas duration is a timedelta.

But beside that, I think they both have the same purpose : set a stop_time for the last stop of the trip.

grote:
> both have the same purpose : set a stop_time for the last stop of the trip.

Maybe that's a reason to unify both properties into one?

jamescr:
I know you don't care how data is structure within OSM, but [`travel_time`](https://taginfo.openstreetmap.org/keys/travel_time#combinations) is more often used in combination with `public_transport:version` than [`duration`](https://taginfo.openstreetmap.org/keys/duration#combinations).  😉

grote:
Wohoo nice new photo @jamescr! 😎 

I'd be fine consolidating things to `travel_time`. We would just need to harmonize the data type across the creators.

jamescr:
> Wohoo nice new photo @jamescr! 😎

Thanks

xamanu:
Yes, indeed very nice photo, @jamescr! You look a bit different, than the last time I saw you :)

The OpenStreetMap terminology clearly seems to prefer `duration`:

* https://wiki.openstreetmap.org/wiki/Key:duration
* https://wiki.openstreetmap.org/wiki/Tag:route%3Dbus

Not even one mention of `travel_time` in the wiki.

jamescr:
> Yes, indeed very nice photo, @jamescr! You look a bit different, than the last time I saw you :)

Thanks, I'm getting old ;-)

> The OpenStreetMap terminology clearly seems to prefer duration:
> *   https://wiki.openstreetmap.org/wiki/Key:duration
>  *  https://wiki.openstreetmap.org/wiki/Tag:route%3Dbus

I did wrong a search on the wiki [here](https://wiki.openstreetmap.org/wiki/Proposed_features/Public_Transport#Route) and [there](https://wiki.openstreetmap.org/wiki/Relation:route#Tags).

Thanks for the correction, is really nice to have you back collaborating.

xamanu:
[...]

PR #96, opened 2017-12-07
WIP: Data structure 2.0

xamanu:
This is a new pull request for the overhauled data structure #30 with a lot of wonderful work by @AltNico
It is similar to #89 (closing now), but rebased on #91, which seems to be the next to get in.


xamanu:
The CI pipeline will fail, because of an error risen by pylint is related to attr.s and can not be fixed: python-attrs/attrs#215

xamanu:
@nlehuby and @prhod, may I request your help on this one? This is a wonderful but deep change and I think it will be difficult to finalize it without your support.

**Status on Accra:**

* It builds (yeah!)
* Accra's unit tests fail because of `File "osm2gtfs/tests/accra_unittests.py", line 151, in test_gtfs_from_cache 'The generated GTFS is different from the expected one')`
* Indeed the GTFS provided for the tests (`accra_tests.zip.ref`) is 1.1M, while the generated one (`accra_tests.zip`) is almost half: 602K

**Status on Fenix:**

* Works until Overpass API complains. No chance to test it further.

**Status on Nicaragua:**

* We use this structure already to generate our GTFSes, but in a different branch without any rebasing. Needs to be further tested.

xamanu:
I changed the cache file names a bit, so this PR can coexists with caching with the old version (before this PR). No need to delete cached files.

xamanu:
In order to make Fenix working properly and to improve the data structure the use of `travel_time` and `duration` was unified and generally improved because all kind of tags of the routes/itineraries are now available in the creators (#76).

osm2gtfs/core/routes.py
```diff
@@ -0,0 +1,117 @@
...
+    _itineraries = attr.ib(default=attr.Factory(list))
+
+    def __attrs_post_init__(self):
+        '''
+        Populates the object with information obtained from the tags
+        '''
+        self.name = self.tags['name']
+
+        if "colour" in self.tags:
+            self.route_color = OsmConnector.get_hex_code_for_color(
```
nlehuby: you need to import OsmConnector from `osm2gtfs.core.osm_connector`

osm2gtfs/creators/accra/trips_creator_accra.py
```diff
@@ -18,71 +19,83 @@ def add_trips_to_schedule(self, schedule, data):
...
 
             route_index = 0
-            for a_route_ref, a_route in line.routes.iteritems():
-                trip_gtfs = line_gtfs.AddTrip(schedule)
+            itineraries = line.get_itineraries()
+            for a_route in itineraries:
+                trip_gtfs = line_gtfs.AddTrip(feed)
                 trip_gtfs.shape_id = TripsCreator.add_shape(
-                    schedule, a_route_ref, a_route)
+                    feed, a_route.route_id, a_route)
```
nlehuby: you need to replace `a_route.route.id` by `a_route.osm_id` to keep the shape_id identical.

osm2gtfs/tests/accra_unittests.py
```diff
@@ -100,8 +100,8 @@ def test_refresh_routes_cache(self):
         self.assertEqual(len(routes), 277, 'Wrong count of routes in the cache file')
```
nlehuby: you need to change the name of the cache file here too (`routes-accra` to `accra-routes`)

osm2gtfs/tests/accra_unittests.py
```diff
@@ -117,32 +117,32 @@ def test_refresh_stops_cache(self):
```
nlehuby: same here

xamanu:
Thanks for the review, @nlehuby. I adjusted the things you mentioned. The whole state is currently still moveing in things from the other branch and make it fit together. Probably you want to wait, until we got it to a stage where we are more confident asking you for help again.

xamanu:
I will close this PR, as it is still not ready, yet, and I don't want to spam you (more) with push notifications. We will re-open, once we are in a shape that we would kindly ask for a review. Thanks.

In case you want to support the ongoing development please have a look on this branch: https://github.com/mapanica/osm2gtfs/tree/data-structure-rebased

Issue #104, opened 2017-12-16
Color handling


xamanu:
Transit lines in GTFS have a `route_color` and a `route_text_color`, whereas the text color "[should provide sufficient contrast when viewed on a black and white screen](https://developers.google.com/transit/gtfs/reference/)".

osm2gtfs tries to get this information from OpenStreetMap's [`colour`](http://wiki.openstreetmap.org/wiki/Key:colour) tag. This can be information in RB hex code or a W3C color name.

In a first step, in #99, it is introduced to map the information from `colour` to the `Line` object to be used in the creators. The standard creator assume RGB hex colors, as also desired by GTFS. However for Estelí two good and additional functionalities are used, that could be considered to be moved also into the core of osm2gtfs:

* Transform W3C color code into RGB hex code.
* Calculate a suitable `route_text_color` based on `route_color`.

xamanu:
In the current Estelí implementation the two things are addressed the following:

* Transform W3C color code into RGB hex code: The python extension [webcolors](https://webcolors.readthedocs.io) is used to provide the functionality. A proper test of there is a valid name of a valid RGB hex code or a non-valid string would need to be included, I guess, in order to make it attractive to the core.
* Calculate a suitable `route_text_color` based on `route_color`: Currently it uses this [code snippet](https://stackoverflow.com/questions/38478409/finding-out-complementary-opposite-color-of-a-given-color/38478744#38478744) which calculates the complementary or opposite color in the color spectrum. As far as I understand, this doesn't seem to fit the requirement GTFS is putting on this, to have a good contrast. I think we need more a function that checks the "brightness" of the given `route_color` and returns either `#000000` (black) or `#ffffff` (white).

xamanu:
I think those screenshots show pretty nicely why a white-black decision (right side) should be taken to generate the `route_text_color` rather than the complementary color (left side):

![image](https://user-images.githubusercontent.com/97706/34260762-367b5030-e667-11e7-86b1-6a47330ecbea.png)

ialokim:
> I think those screenshots show pretty nicely why a white-black decision (right side) should be taken to generate the route_text_color rather than the complementary color (left side)

You're completely right 👍  Looks like ugly the left side 🤢😅

> I think we need more a function that checks the "brightness" of the given route_color and returns either #000000 (black) or #ffffff (white).

[...]

PR #99, opened 2017-12-14
Data structure, standard creators, Nicaragua networks

xamanu:
This is a new pull request for the overhauled data structure, now with standard creators and Nicaragua config files that rely on these. It comes with a lot of wonderful work by @AltNico and @ialokim.

Following a request of @ialokim, I'm opening a PR already. It is meant to be subject for review by all people involved in osm2gtfs.

This PR addresses the following issues:

* #30 - Data structure
* #26 - Default implementation for route creation
* #33 - Default implementation for trips creation
* #13 - Get Route Information from OSM
* #71 - Inconsistent usage of a parameter in Route Class
* #76 - Unify travel_time and duration

PR #114, opened 2017-12-23
Improve color handling

xamanu:
This is the implementation of #104 to improve color handling and move essential parts into the core part of osm2gtfs. This PR includes:

* `Line` objects' `route_color` attributes are now a hex value (and shouldn't be different things).
* Data from OSM with `colour` information is sanitized (checked and eventually converted from W3C color names to hex values).
* The `route_text_color` can be obtained from OSM by using the `ref:colour_tx` tag. This seems to be the most appropriate, when looking at the [wiki](https://wiki.openstreetmap.org/wiki/Key:ref:colour). Of course creators are free to use any other tag.
* If no information on the text color has been found in OSM; the standard `RoutesCreator` computes it's own `route_text_color`, based on a [smart logic](http://www.nbdtech.com/Blog/archive/2008/04/27/Calculating-the-Perceived-Brightness-of-a-Color.aspx) found by @ialokim and implemented within the helper function `calculate_color_of_contrast`. Of course creators can override this standard creator behaviour to define the colors and the logic as they want/need.

This PR is based on #99 (data structure).


xamanu:
Rebased on latest upstream master branch

ialokim:
I've only rebased on current master and squashed the two commits into one. I think this is really ready to be merged!

Issue #97, opened 2017-12-09
Improve standard stop creator's station building process


xamanu:
There are very interesting comments about the standard stop creators, mentioned by @nlehuby and @prhod in the PR #66:

There are different possible scenarios on stop creation regarding "[parent_stations](https://developers.google.com/transit/gtfs/reference/#stop)" (in GTFS) or "[stop_area](http://wiki.openstreetmap.org/wiki/Tag:public_transport%3Dstop_area)" (in OSM):

> * you can create GTFS stops without any grouping [old behavior]
> * you can group GTFS stops using OSM StopArea [current behavior]
> * you can group GTFS stops by name and proximity [Accra creator)]
> * you can mix it all (using OSM Stop Area and grouping them by name and proximity)

And there were some great ideas on maybe mixing them in the standard creator, or find a way how the proximity search can become part of the general code and not exclusive on Accra.

Let's use this issue to develop further on this.

xamanu:
As a first step I would suggest to move the function `get_crow_fly_distance` into the newly introduced `Helper` class. Then it can be easily included and reused by other creators.

ialokim:
Related to this is proper name handling: What happens if the OSM `stop_position` and `platform` have a different name than the OSM `stop_area`? I'm thinking of the way we tagged bus station like [this one](osm.org/relation/7637880): Only the `stop_area` presents the "proper" name, the `platform`s each reflect the destinations of the buses that stop on this `stop_position`.

Should all the GTFS stops have the OSM `stop_area`'s name or only the GTFS station? @grote What is rendered e.g. in Transportr, GTFS stop or GTFS station names?

grote:
Navitia needs stop areas for each stop point. If none exists, it creates one on its own. Both can then be shown in Transportr. The auto-completion shows the stop area, but in trips, individual stop points can be shown, so both should have names.

xamanu:
I'm curious: does Navitia need OSM `stop_areas` or GTFS "Stations" for each stop?

grote:
Their debug output talks about stop areas, but I guess they mean GTFS "Stations". The people working on Navitia can probably clarify.

prhod:
Hi, 
there are some cases where the stop_point and the corresponding stop_area don't have the same name, but in general the name is the same.
[...]

PR #115, opened 2017-12-23
Include distance calculation in Helper

xamanu:
This implements a first step for #97 and moves the great function `get_crow_fly_distance` into the newly introduced `Helper` class. Then it can be easily included and reused by other creators.

This PR is based on #99 (data structure).


xamanu:
Rebased on latest upstream master branch

PR #119, opened 2018-01-13
Bug generating feed info end date automatically

jamescr:
This PR fix a bug when generating the feed info `end date` automatically for the January special case.


grote:
Looks good. Thanks for catching this!

PR #120, opened 2018-01-20
unifies object type after reading the schedule source files

jamescr:
In configuration.py the `schedule_source` variable type varies depending if it is a URL or a local file. The former is a file-like and the latter a string. The write_file method in cache.py assumes a file object and tries to read() from the content param, this causes an error when the schedule_source of the configuration file is a local file (a str). With this PR the type of the `schedule_source` is str in both cases.

### DIFF
--- a/osm2gtfs/core/configuration.py
+++ b/osm2gtfs/core/configuration.py
@@ -81,7 +81,7 @@ class Configuration(object):
                     sys.stderr.write(
                         "Error: Couldn't find schedule_source file.\n")
                     sys.exit(0)
-                schedule_source = schedule_source_file
+                schedule_source = schedule_source_file.read()
 
         # Cache data
         Cache.write_file(cached_file, schedule_source)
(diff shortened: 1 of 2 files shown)


osm2gtfs/core/configuration.py
```diff
@@ -76,7 +76,7 @@ def get_schedule_source(self, refresh=False):
             else:
                 # Check if it is a valid url
                 try:
-                    schedule_source_file = urlopen(source_file)
+                    schedule_source_file = urlopen(source_file).read()
```
ialokim: I think the `read()` method should be called in [line 84](#diff-07e4c70d96a2beb51112609727f78897R84) to keep it kind of consistent with [line 74](#diff-07e4c70d96a2beb51112609727f78897R74).

jamescr: yes, you are right. I'll changed it.

grote:
Alright, by popular demand, I am merging this before #99. Thanks @jamescr!

PR #122, opened 2018-01-27
Upgrade to last Transitfeed

prhod:


grote:
Wow, that is strange! Suddenly everything works!???

However, before upgrading, we should remove the feed-info workaround at least.

prhod:
The actual test only check for the size of the file, so there could still be something.
I'm adding the `do not merge` label until I add a test on the OSM relation I noticed before. Sorry for the too early PR !

prhod:
I added a manual diff on stop_times (to have more explicit messages) and found the bug. It was not related to the upgrade of TransitFeed !
cheers to @nlehuby who told me what the bug could be by reading the code for 10 minutes when I struggled for hours ;)

grote:
Thanks @prhod!

Could you please also remove the feed_info workaround for this to get merged? It is essentially just reverting 7101a1f3f07629eabe3b77ac286bc5baf1c4572a.

prhod:
Sorry for the delay, I made the revert :)

PR #123, opened 2018-02-12
 Support multiple values for the same OSM tag

### LINKED ISSUE #110: Support multiple values for the same OSM tag
Maybe this is already supported yet, but in order to be able to [generate national GTFS feeds](https://github.com/grote/osm2gtfs/milestone/1) it is important to be able to specify multiple values for just one OSM tag. This should result in queries that append the different values separated by an *or*.

Example: 
```json
{
    "query": {
        "tags": {
            "network": "NI",
            "network": "NI-Estelí",
            "network": "NI-Managua"
        }
    }
}

### PR DESCRIPTION
**Pull request by ialokim**
I've elaborated this little code change to support a `config.json` file as follows:
```json
{
    "query": {
        "tags": {
            "public_transport:version": "2",
            "operator":  [
                "Incofer",
                "Alpizar S.A."
            ]
        }
    }
}
```

This generated the following Overpass query:
```qml
relation
  (9.8346,-84.2418,10.0365,-83.8507)
  ["public_transport:version" = "2"]
  ["operator" ~ "Incofer|Alpizar S.A."];
(._;);
out body;
```

Fixes #110

### DIFF
--- a/osm2gtfs/core/osm_connector.py
+++ b/osm2gtfs/core/osm_connector.py
@@ -42,7 +42,11 @@ class OsmConnector(object):
         # tags from config file for querying
         self.tags = ''
         for key, value in self.config["query"].get("tags", {}).iteritems():
-            self.tags += unicode('["' + key + '" = "' + value + '"]')
+            if isinstance(value, list):
+                value = '^' + '$|^'.join(value) + '$'
+                self.tags += unicode('["' + key + '" ~ "' + value + '"]')
+            else:
+                self.tags += unicode('["' + key + '" = "' + value + '"]')
         if not self.tags:
             # fallback
             self.tags = '["public_transport:version" = "2"]'


xamanu:
Looks good to me. Have you tried it with special characters, like `Estelí`?

ialokim:
> Have you tried it with special characters, like Estelí?

It should work as before, the unicode handling isn't changed by this PR. So yes, special characters should be supported as before.

osm2gtfs/core/osm_connector.py
```diff
@@ -42,7 +42,11 @@ def __init__(self, config):
         # tags from config file for querying
         self.tags = ''
         for key, value in self.config["query"].get("tags", {}).iteritems():
-            self.tags += unicode('["' + key + '" = "' + value + '"]')
+            if isinstance(value, list):
+                value = '|'.join(value)
+                self.tags += unicode('["' + key + '" ~ "' + value + '"]')
```
nlehuby: Why do you switch to `~` ?

It would work with a `=` here too, and would be safer : we don't want Overpass to match regular expression and get matches that partially fit the values.

ialokim: > ...and get matches that partially fit the values.

Well, I've simply followed the suggestion by @jamescr in [#110](https://github.com/grote/osm2gtfs/issues/110#issuecomment-353234482), but you are right, that's indeed a problem.

I'll look into it.

ialokim: I've changed it accordingly and added regex anchors (`^` and `$`) to prevent partially matching.

ialokim: Thanks a lot for noticing it!

PR #124, opened 2018-02-14
Bugfix: check for NoneType in OSM Connector

ialokim:
In case of invalid osm objects, `OSMConnector` returns None, but always expects a valid `Line` or `Itinerary` object. This PR should fix this.

### DIFF
--- a/osm2gtfs/core/osm_connector.py
+++ b/osm2gtfs/core/osm_connector.py
@@ -134,8 +134,9 @@ class OsmConnector(object):
                 # Create Itinerary objects from member route variants
                 if member.ref in route_variants:
                     rv = route_variants.pop(member.ref)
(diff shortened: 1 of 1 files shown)


xamanu:
Maybe it would be good to get a suitable error or warning message. In which cases has this happened?

ialokim:
> Maybe it would be good to get a suitable error or warning message. In which cases has this happened?

The subfunctions already print these warning messages out, see line [365](/mapanica/osm2gtfs/blob/d1e9ca2db27f8f06681baf96405391004be813e2/osm2gtfs/core/osm_connector.py#L365) and line [296](/mapanica/osm2gtfs/blob/d1e9ca2db27f8f06681baf96405391004be813e2/osm2gtfs/core/osm_connector.py#L365). So there's no need to add another message, only to handle the `NoneType` correctly.

PR #125, opened 2018-02-15
Fix encoding errors

ialokim:
While working on the countrywide NicaraguaCreator, I've found some odd python 2 encoding errors:
* `transitfeed`'s validator expects `route_long_name` and `route_short_name` to have the same encoding. That's why I changed the `RoutesCreator`.
* `str` expects only ASCII values, if the `route_id` contains non-ASCII chars, it throws an error. These are the changes to `TripsCreator`.

This PR should fix them both.


grote:
Thanks for spotting and the fix!

PR #126, opened 2018-02-15
Fixed some functions in StandardTripsCreator

ialokim:
While working on the countrywide NicaraguaCreator, I've found some issues with the current `TripsCreator` implementation. See the inline comments for the specific issues.

### DIFF
--- a/osm2gtfs/creators/trips_creator.py
+++ b/osm2gtfs/creators/trips_creator.py
@@ -87,7 +87,8 @@ class TripsCreator(object):
             if input_fr == itinerary.fr and input_to == itinerary.to:
                 trip_services = trip["services"]
                 for service in trip_services:
-                    services.append(service)
+                    if service not in services:
+                        services.append(service)
 
         if not services:
             print(" Warning: From and to values didn't match with schedule.")
@@ -124,13 +125,31 @@ class TripsCreator(object):
                   str(line.route_id) + ")")
             print(" " + itinerary.osm_url)
             print(" " + line.osm_url)
-            return True
+            return False
 
         # Check if time information in schedule can be found for
         # the itinerary
         if itinerary.route_id not in schedule['lines']:
-            print(" Warning: Route not found in schedule.")
+            print("Warning: Route not found in schedule.")
             return False
+
+        # Check if from and to tags are valid and correspond to
+        # the actual name of the first and last stop of the itinerary.
+        for trip in schedule['lines'][itinerary.route_id]:
+            if (trip["from"] == itinerary.fr and trip["to"] == itinerary.to):
+                trip_stations = trip["stations"]
+                if trip_stations[0] != itinerary.fr:
+                    print("Warning: The first station in the stations list of trip (" +
+                          str(itinerary.route_id) + ") doesn't match first station of itinerary.")
+                    print(" " + itinerary.osm_url)
+                    print(" Please review the schedule file.")
+                    return False
+                elif trip_stations[-1] != itinerary.to:
+                    print("Warning: The last station in the stations list of trip (" +
+                          str(itinerary.route_id) + ") doesn't match last station of itinerary.")
+                    print(" " + itinerary.osm_url)
+                    print(" Please review the schedule file.")
+                    return False
 
         return True
 
@@ -165,16 +184,16 @@ class TripsCreator(object):
             gtfs_trip = route.AddTrip(feed, headsign=itinerary.to,
                                       service_period=trip_builder['service_period'])
             trips_count += 1
+            search_idx = 0
 
             # Go through all stops of an itinerary
-            for itinerary_stop_id in itinerary.get_stops():
+            for itinerary_stop_idx, itinerary_stop_id in enumerate(itinerary.get_stops()):
 
                 # Load full stop object
                 try:
                     itinerary_stop = trip_builder[
                         'all_stops']['regular'][itinerary_stop_id]
                 except ValueError:
-
                     sys.stderr.write(
                         "Itinerary (" + itinerary.route_url + ") misses a stop: \n")
                     sys.stderr.write(
@@ -187,20 +206,40 @@ class TripsCreator(object):
                 except ValueError:
                     print("Warning: Stop in itinerary was not found in GTFS.")
                     print(" " + itinerary_stop.osm_url)
+                    continue
 
                 # Make sure we compare same unicode encoding
                 if type(itinerary_stop.name) is str:
                     itinerary_stop.name = itinerary_stop.name.decode('utf-8')
 
-                time = "-"
+                schedule_stop_idx = -1
                 # Check if we have specific time information for this stop.
                 try:
-                    time = trip[trip_builder['stops'].index(itinerary_stop.name)]
+                    schedule_stop_idx = trip_builder['stops'].index(itinerary_stop.name, search_idx)
                 except ValueError:
-                    pass
+                    if itinerary_stop.get_parent_station() is not None:
+                        # If stop name not found, check for the parent_station name, too.
+                        itinerary_station = trip_builder[
+                            'all_stops']['stations'][str(itinerary_stop.get_parent_station())]
+                        if type(itinerary_station.name) is str:
+                            itinerary_station.name = itinerary_station.name.decode('utf-8')
+                        try:
+                            schedule_stop_idx = trip_builder[
+                                'stops'].index(itinerary_station.name, search_idx)
+                        except ValueError:
+                            pass
 
-                # Validate time information
-                if time != "-":
+                # Make sure the last stop of itinerary will keep being the last stop in GTFS
+                last_stop_schedule = schedule_stop_idx == len(trip_builder['stops']) - 1
+                last_stop_itinerary = itinerary_stop_idx == len(itinerary.get_stops()) - 1
+                if last_stop_schedule != last_stop_itinerary:
+                    schedule_stop_idx = -1
+
+                if schedule_stop_idx != -1:
+                    time = trip[schedule_stop_idx]
+                    search_idx = schedule_stop_idx + 1
+
+                    # Validate time information
                     try:
                         time_at_stop = str(
                             datetime.strptime(time, "%H:%M").time())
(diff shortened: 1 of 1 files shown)


osm2gtfs/creators/trips_creator.py
```diff
@@ -88,7 +88,8 @@ def _prepare_trips(self, feed, schedule, itinerary):
             if input_fr == itinerary.fr and input_to == itinerary.to:
                 trip_services = trip["services"]
                 for service in trip_services:
-                    services.append(service)
+                    if service not in services:
```
ialokim: To prevent the case when the `itineraries` in the `Schedule` are split by some reason, do not double the `services`.

osm2gtfs/creators/trips_creator.py
```diff
@@ -125,14 +126,31 @@ def _verify_data(self, schedule, line, itinerary):
                   str(line.route_id) + ")")
             print(" " + itinerary.osm_url)
             print(" " + line.osm_url)
-            return True
+            return False
```
ialokim: I'm not 100% sure, but shouldn't it return `False` when there's an error?

osm2gtfs/creators/trips_creator.py
```diff
@@ -125,14 +126,31 @@ def _verify_data(self, schedule, line, itinerary):
...
 
         # Check if time information in schedule can be found for
         # the itinerary
         if itinerary.route_id not in schedule['lines']:
-            print(" Warning: Route not found in schedule.")
+            print("Warning: Route not found in schedule.")
```
ialokim: Just to keep it as in the other cases.

osm2gtfs/creators/trips_creator.py
```diff
@@ -125,14 +126,31 @@ def _verify_data(self, schedule, line, itinerary):
...
             return False
 
+        # Check if from and to tags are valid and same one as first and last itinerary.stop
```
ialokim: This explains well the added check.

osm2gtfs/creators/trips_creator.py
```diff
@@ -166,16 +184,16 @@ def _add_itinerary_trips(self, feed, itinerary, line, trip_builder,
             gtfs_trip = route.AddTrip(feed, headsign=itinerary.to,
                                       service_period=trip_builder['service_period'])
             trips_count += 1
+            search_idx = 0
```
ialokim: To prevent problems in `itineraries` where a stop (or a stop_name) is served twice, so that it doesn't pick the same time information twice (this would give "high-speed services" for the transitfeed validator), I've introduced the index from where it should search for the next time.

osm2gtfs/creators/trips_creator.py
```diff
@@ -188,20 +206,40 @@ def _add_itinerary_trips(self, feed, itinerary, line, trip_builder,
...
                 try:
-                    time = trip[trip_builder['stops'].index(itinerary_stop.name)]
+                    schedule_stop_idx = trip_builder['stops'].index(itinerary_stop.name, search_idx)
                 except ValueError:
-                    pass
-
-                # Validate time information
-                if time != "-":
+                    if itinerary_stop.get_parent_station() is not None:
+                        # If stop name not found, check for the parent_station name, too.
```
ialokim: Here it is the explanation.

osm2gtfs/creators/trips_creator.py
```diff
@@ -188,20 +206,40 @@ def _add_itinerary_trips(self, feed, itinerary, line, trip_builder,
...
+                            'all_stops']['stations'][str(itinerary_stop.get_parent_station())]
+                        if type(itinerary_station.name) is str:
+                            itinerary_station.name = itinerary_station.name.decode('utf-8')
+                        try:
+                            schedule_stop_idx = trip_builder[
+                                'stops'].index(itinerary_station.name, search_idx)
+                        except ValueError:
+                            pass
+
+                # Make sure the last stop of itinerary will keep being the last stop in GTFS
```
ialokim: To solve same problem described above.

osm2gtfs/creators/trips_creator.py
```diff
@@ -217,10 +255,10 @@ def _add_itinerary_trips(self, feed, itinerary, line, trip_builder,
                 else:
                     try:
                         gtfs_trip.AddStopTime(gtfs_stop)
-                    except ValueError:
-                        print("Warning: Could not add first a stop to trip.")
+                    except transitfeed.problems.OtherProblem:
```
ialokim: It never throws `ValueError`, but `OtherProblem`.

osm2gtfs/creators/trips_creator.py
```diff
@@ -217,10 +255,10 @@ def _add_itinerary_trips(self, feed, itinerary, line, trip_builder,
...
+                        print("Warning: Could not add first stop to trip without time information.")
                         print(" " + itinerary_stop.name +
-                              " - " + itinerary_stop.osm_id)
+                              " - " + itinerary_stop.osm_url)
```
ialokim: just to give more and consistent information

osm2gtfs/creators/trips_creator.py
```diff
@@ -297,13 +335,14 @@ def _load_itinerary_schedule(self, schedule, itinerary, service):
...
-        times = None
+        times = []
         for trip in schedule['lines'][itinerary.route_id]:
             trip_services = trip["services"]
             if (trip["from"] == itinerary.fr and
                     trip["to"] == itinerary.to and
                     service in trip_services):
-                times = trip["times"]
+                for time in trip["times"]:
+                    times.append(time)
```
ialokim: In order to not loose times (overriding them).

osm2gtfs/creators/trips_creator.py
```diff
@@ -323,4 +362,5 @@ def _load_scheduled_stops(self, schedule, itinerary, service):
                         "to"] == itinerary.to and service in trip_services):
                 for stop in trip["stations"]:
                     stops.append(stop)
+                break
```
ialokim: In order to not duplicate stops.

osm2gtfs/creators/trips_creator.py
xamanu: I tested with Managua and it is the same result. Generally checked the code and looks good.

osm2gtfs/creators/trips_creator.py
ialokim: They are in some cases of the [Nicaraguan `timetable.json`](/mapanica/pt-data-nicaragua/blob/master/timetable.json) because of the generation by [easy-timetable-generator](/mapanica/easy-timetable-generator). But anyway, why shouldn't it be supported?

osm2gtfs/creators/trips_creator.py
ialokim: Changed accordingly.

ialokim:
> Tested it with Managua

As Estelí and Managua are currently the only regions that use the standard creator's implementation (please correct me if I am wrong), I think this is ready to be merged too.

osm2gtfs/creators/trips_creator.py
xamanu: I still don't understand it. Can you explain concretely why they should be split. Sorry for asking this stupid question, I just didn't get it, yet.

ialokim: I'm referring to the case when there are two or more dict items inside the line's list for the same line, which present different times but using the same service period. See this example:

```json
    "lines": {
        "E-MAN-EST": [
            {
                "exeptions": [],
                "from": "Mercado Mayoreo",
                "services": [
                    "Mo-Su"
                ],
                "stations": [
                    "Mercado Mayoreo",
                    "COTRAN Sur"
                ],
                "times": [
                    [
                        "05:45",
                        "08:15"
                    ],
                    [
                        "08:15",
                        "10:45"
                    ],
                    [
                        "09:15",
                        "11:45"
                    ],
                    [
                        "10:45",
                        "13:15"
                    ],
                    [
                        "11:45",
                        "14:15"
                    ]
                ],
                "to": "COTRAN Sur"
            },
            {
                "exeptions": [],
                "from": "Mercado Mayoreo",
                "services": [
                    "Mo-Su"
                ],
                "stations": [
                    "Mercado Mayoreo",
                    "COTRAN Sur"
                ],
                "times": [
                    [
                        "13:15",
                        "15:45"
                    ],
                    [
                        "15:15",
                        "17:45"
                    ],
                    [
                        "13:45",
                        "16:15"
                    ],
                    [
                        "15:45",
                        "18:15"
                    ]
                ],
                "to": "COTRAN Sur"
            },
            ...
```

Skippern:
I don’t think multiple dicts like this should have a negative impact on the end result.

Aun Johnsen

osm2gtfs/creators/trips_creator.py
xamanu: Thanks for the explanation. I think this (edge) case is kind of an inconsistency in the standard schedule format and I personally would rather require a clean input format. But I guess if you and @Skippern are in favour of supporting it, there is a deeper reason I still can't see.

xamanu: But shouldn't then the two dicts not been merged? As far as I see, now the second one (and all that follow) would be just discarded.

ialokim: > now the second one (and all that follow) would be just discarded

To clear it up a bit, I'll explain what happened before and what should happen IMHO:
* Before this PR, the `services` variable contained the following after iterating over the two (or more) dicts for one line: `[Mo-Su, Mo-Su]`. So in line [98](https://github.com/mapanica/osm2gtfs/blob/32524f075858f746dad0353620c900d51beb56ad/osm2gtfs/creators/trips_creator.py#L98) and more specifically in line 112, it would append the times from both dicts __twice__ to the `trips` variable with the same service period `Mo-Su`.
* With this change, the `service` variable would only contain the following: `[Mo-Su]` and the trips would be getting doubled.

This code isn't about the whole dict that would be discarded, but only preventing the doubling of a service period.

xamanu: Now I got it. Thanks!

grote:
Maybe let's get reviews from others whose creators rely on this?

xamanu:
So far I think only the Nicaraguan providers rely on the standard trips creator.

Issue #118, opened 2017-12-27
Enabling a fast simple creator without timetable


prhod:
I think we could use a simplification of the creation of a new creator by allowing the generation of a GTFS without any `trips`, `calendar` or `stop_times` with the default creator and only a config file. (I think @nlehuby talked about that with @xamanu in a comment a few days ago).
I think it could be usefull for new users of osm2gtfs when working  on a new creator for a city, or just to extract OSM Transit data for simple analysis in QGIS.

nlehuby:
I think it would be even better if it creates a valid GTFS, with all files, but with dumb schedules (one trip per route & +1min between each stops, or so).

The `trips` and `stop_times` are interesting too where working on a new creator, because they list the stops in the right order for each route.

nlehuby:
We could have this by implementing #146 and add dumb default values (if not set in OSM). 
For instance:
* interval = 12:00
* opening_hours = Mo 08:00-20:00
* duration = 01:00

xamanu:
I like the idea. We just should pay attention to not remove important functionality. IMHO we would need a way to select the different standard creators from the command line. There I could think about different solutions (let the user select, select on the availability of input etc).

PR #127, opened 2018-02-15
Remove unused stops and routes

ialokim:
This PR adds the possibility to remove unused stops and routes from the final _GTFS_ file. In order to allow #118, there could be easily added an argument switch.

### DIFF
--- a/osm2gtfs/creators/stops_creator.py
+++ b/osm2gtfs/creators/stops_creator.py
@@ -41,6 +41,24 @@ class StopsCreator(object):
             # Add stop to feed
             gtfs_stop_id = self._add_stop_to_feed(stop, feed)
 
+    def remove_unused_stops_from_feed(self, feed):
+        """
+        This function removes every stop which is not used by any trip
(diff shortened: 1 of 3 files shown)


osm2gtfs/creators/stops_creator.py
```diff
@@ -41,6 +41,19 @@ def add_stops_to_feed(self, feed, data):
             # Add stop to feed
             gtfs_stop_id = self._add_stop_to_feed(stop, feed)
 
+    def remove_unused_stops_from_feed(self, feed):
```
grote: Could you maybe add a docstring here and above that explains when this method is called?

ialokim: Done.

ialokim:
> Looks like it doesn't add the possibility, but removes the stops and routes right way.

That's right, it removes them right away. But why would one prefer to have them blowing up the filesize? But as stated above, for #118 I could introduce another argument switch to turn this behavior of removing off?

PR #128, opened 2018-02-15
Add a new command argument to refresh all data

ialokim:
As of the current `master` branch, the command argument `--refresh-all` is misleading, because it only refreshes OSM data but not the schedule information. This PR introduces a new `--refresh-osm` switch to refresh only OSM data and changes the `--refresh-all` switch to renew the time information, too.


xamanu:
Very nice! Looks very good.

grote:
Thanks @ialokim !

PR #129, opened 2018-02-15
Add support for required via tag (if given inside OSM data)

### LINKED ISSUE #101: Support the 'via' tag in the default trip creator
As pointed out [in this pull request](https://github.com/grote/osm2gtfs/pull/99#discussion_r156994991), it would be nice to support the `via` tag in addition to the already supported (and required) `from` and `to` tags. See [the wiki](https://github.com/grote/osm2gtfs/wiki/Schedule#via) for definition of how it should be implemented.

ialokim:
This PR adds support for the required `via` tag if it is set in the OSM data.

After merging in #126, this PR should be rebased and we should change the verification check introduced there by adding the `via` tag.

Fixes #101.


xamanu:
I was first hesitating about the `via` tag, as it [doesn't seem to be an official tag](https://wiki.openstreetmap.org/wiki/Key:via) in OpenStreetMap. However it is [mentioned in the wiki](https://wiki.openstreetmap.org/wiki/Public_transport), part of the [approved proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Public_Transport), and it is [used quite extensively](https://taginfo.openstreetmap.org/keys/via). So, all good, to include it into the osm2gtfs core. It looks good so far. Will wait until #126 got in until I test it thoroughly.

ialokim:
I've only rebased and made pep8 happy :laughing:

osm2gtfs/creators/trips_creator.py
```diff
@@ -317,9 +323,12 @@ def _load_scheduled_stops(self, schedule, itinerary, service):
         stops = []
         for trip in schedule['lines'][itinerary.route_id]:
             trip_services = trip["services"]
+            if "via" not in trip:
+                trip["via"] = None
             if (trip[
                     "from"] == itinerary.fr and trip[
-                        "to"] == itinerary.to and service in trip_services):
+                        "to"] == itinerary.to and trip[
+                        "via"] == itinerary.via and service in trip_services):
```
grote: It was already like this, but breaking open the accessor like this looks like bad style. Can we keep `trip["via"]` together without line breaks?

ialokim:
Before merging this, we should check all the creators' _GTFSes_, because of the new required information in the schedule file. I [had to fix](https://github.com/mapanica/pt-data-esteli/commit/f1bb44e37e37aee94ca2dfbd223949b0530c962c) Esteli's `schedule` accordingly.

xamanu:
I guess the tests (#133, #135, #136 and #137) will be very useful for this.

PR #130, opened 2018-02-16
Add support for national buses of Nicaragua

ialokim:
This PR only adds the minimal `config.json` used by the standard creators. However, to successfully produce the [countrywide GTFS](https://github.com/mapanica/datos/raw/master/ni-gtfs.zip), it depends on #114, #124, #125, #126 and (to clean the data up) on #127.

Those PRs should be merged first.


xamanu:
Yeah, nice progress here! I would like to see #83 going in before also. This list of providers is getting long now.

ialokim:
I've changed the configuration to allow ferries, too (simply removed `route=bus`). So now, the ferries from and to Ometepe (e.g. [this one](https://www.openstreetmap.org/relation/2020594)) are available inside the resulting _GTFS_.

xamanu:
This PR needs to be rebased on the latest commits to the code and modified to stick to the newly introduced naming conventions (see #83 for more information) for selectors.

ialokim:
> This PR needs to be rebased on the latest commits to the code and modified to stick to the newly introduced naming conventions (see #83 for more information) for selectors.

Done. I've renamed the selector to `ni_national` as suggested in [#83](https://github.com/grote/osm2gtfs/pull/83#issuecomment-347992823).

xamanu:
I suggest to include tests for this provider first. Based on the structure of #132.

xamanu:
Yes, indeed, In my opinion every (new) creator should now include tests.

PR #131, opened 2018-02-28
Acceptance tests for Managua, Nicaragua creator

xamanu:
Following the work of adding tests for the Accra, Ghana creator by @prhod and @nlehuby, this PR introduces similar acceptance tests for the Managua, Nicaragua creator.


xamanu:
Running the checks with the new tests of Managua the [whole process takes now around 24 minutes](https://travis-ci.org/grote/osm2gtfs/builds/347439268?utm_source=github_status&utm_medium=notification). It is doing practically the same as for the Accra tests, but Managua takes longer. I assume (but don't know exactly) it is because it calculates a schedule, whereas Accra relies on frequencies only.

xamanu:
Rebased on latest `master`.

grote:
Looks like the acceptance tests share lots of code with each other, maybe this can be factored out into a super class?

xamanu:
Yes, basically there are some variables (which files to use for testing, which selector, how many routes and stops are expected to be generated) and the input files (fixtures) that differ. The logic is largely the same.

grote:
Do you want to do this now or in a separate PR?

xamanu:
Both is fine for me. Although tend to suggest to accept this PR and then unifying the creator test logic in a separate PR. This way, we make sure that this refactoring is not leading to an regression itself and it will be easier and better for others to review.

PR #132, opened 2018-02-28
Add check for known OSM route types

ialokim:
The standard trips creator currently only works for bus services, e.g. for `route=train`, it fails currently. This PR adds a dictionary to the `Line` object, that matches the commonly used OSM public transport `route` tag values to the [GTFS names](https://developers.google.com/transit/gtfs/reference/#routestxt).


xamanu:
But please fix the PEP8 error in the checks. Thanks.

ialokim:
> But please fix the PEP8 error in the checks.

Done :)

grote:
Thanks @ialokim for all these awesome contributions!

PR #133, opened 2018-03-01
Abstract test classes to share common logic between creators

xamanu:
As a follow up to #131, this PR proposes to create a class `CreatorsTestsAbstract` with the general testing logic for providers/creators, which was before implemented very similarily in the `gh_accra` and `ni_managua` providers. With this PR, the providers just have to extend from this class and define the necessary variables through overridable methods:

* `_get_selector()` (required)
* `_get_required_variables()` (required)
* `_get_standard_variables()` (optional)
* `_override_configuration()` (optional)

Unfortunately the comparing diff tool seems to be quite overwhelmed by these kind of changes, and it is difficult to see the changes. Looking at the code, without using the diff tool is probably easier.


osm2gtfs/tests/creators/tests_creators_abstract.py
```diff
@@ -0,0 +1,323 @@
...
+            api = overpy.Overpass()
+            mocked1.return_value = api.parse_xml(overpass_xml)
+            data.get_stops(refresh=True)
+        self.assertTrue(os.path.isfile(cache_file), 'The stops cache file creation failed')
+        cache = Cache()
+        stops = cache.read_data(self.selector + "-stops")
+        amount_of_stops = len(stops['regular']) + len(stops['stations'])
+        self.assertEqual(
+            # amount_of_stops, self.count_stops, # WORKS FOR ACCRA
+            amount_of_stops, self.count_stops + self.count_stations, # WORKS FOR MANAGUA
```
xamanu: Here we have an incompatibility between Accra and Managua and I would like to ask for your help on this, @prhod and @nlehuby. For some reason the GTFS for Managua counts stops and stations (stop_areas), whereas Accra only counts the stops and not the stations. However in line [100](https://github.com/xamanu/osm2gtfs/blob/3b4ad11e744a06ee1a36f9dc3c6ba46666cee14c/osm2gtfs/tests/creators/tests_creators_abstract.py#L100) explicitly stops and stations are added up. So I guess Managua is following the correct behaviour. But I'm really not sure on this one and how to fix it for Accra. Do you have an idea?

osm2gtfs/tests/creators/tests_creators_abstract.py
```diff
@@ -0,0 +1,322 @@
...
+            api = overpy.Overpass()
+            mocked1.return_value = api.parse_xml(overpass_xml)
+            data.get_stops(refresh=True)
+        self.assertTrue(os.path.isfile(cache_file), 'The stops cache file creation failed')
+        cache = Cache()
+        stops = cache.read_data(self.selector + "-stops")
+        amount_of_stops = len(stops['regular']) + len(stops['stations'])
+        self.assertEqual(
+            # amount_of_stops, self.count_stops,  # WORKS FOR ACCRA
+            amount_of_stops, self.count_stops + self.count_stations,  # WORKS FOR MANAGUA
```
xamanu: Here we have an incompatibility between Accra and Managua and I would like to ask for your help on this, @prhod and @nlehuby.

For some reason the GTFS for Managua counts stops and stations (stop_areas), whereas Accra only counts the stops and not the stations. However in line [99](https://github.com/xamanu/osm2gtfs/blob/db43c073c0fc73bcf90f973237b17ccfe9402d05/osm2gtfs/tests/creators/tests_creators_abstract.py#L99) explicitly stops and stations are added up. So I guess Managua is following the correct behaviour. But I'm really not sure on this one and how to fix it for Accra.

Do you have an idea?

osm2gtfs/tests/creators/tests_creators_abstract.py
```diff
@@ -0,0 +1,326 @@
...
+
+class TestCreatorsAbstract(unittest.TestCase):
+    def setUp(self):
+        """
+        This method sets the basic variables for testing of providers.
+        It needs to be implemented for each provider in the heriting class.
+        """
+
+        # Define selector for this tests
+        self.selector = "YOUR SELECTOR"
```
grote: Maybe this would be better with an abstract method?

xamanu: This is where my knowledge on Python ends currently. [I gave it a try here](https://github.com/xamanu/osm2gtfs/blob/creator-tests-abstract-playground/osm2gtfs/tests/creators/tests_creators_abstract.py#L37), but it breaks with error message
> TypeError: Can't instantiate abstract class TestCreatorsAbstract with abstract methods setUp

My very wild guess: There seems to be some deep-shit going on because `setUp` is the initiator function.

Can anybody help me or point me to documentation that helps me? Thanks.

grote: You would declare a new method `get_selector()` which throws a NotImplemented exception, so all subclasses need to implement it. Then where you need the selector, you call the method.

xamanu: ok, it is implemented as suggested, with two abstract functions:

* `_get_selector()`
* `_get_required_values()`

And optionally can be overridden:

* `_get_standard_values()`
* `_override_configuration()`

Is this how you meant it, @grote?

osm2gtfs/tests/creators/tests_creators_abstract.py
xamanu: Because of some other changes, the code moved: The important part is now at line 140-[144](https://github.com/xamanu/osm2gtfs/blob/6ebfa85ec86cdc7531daac1c33876422079ed42c/osm2gtfs/tests/creators/tests_creators_abstract.py#L144)

xamanu:
Automatic tests currently fail because of different behaviour on handling stations (stop_areas) between Accra and Managua, as [described here](https://github.com/grote/osm2gtfs/pull/133#discussion_r171654496):

> AssertionError: Wrong count of stops in the cache file

I'd prefer to come to a consistent solution for both providers. Alternatively we would have to add another variable to be set to use in the critical position.

xamanu:
Let's keep this more simple: I introduced a separate variable `stops_osm_count` for this case, where Accra and Managua differ. The reason seems to be: Managua used OSM stop_areas to create stations, while Accra does a proximity search to define stations. I think there is no way out as defining another variable. Or such improvement should happen in a separate PR.

Now, tests work well (it actually fixes an error which lets the current `master` fail).

osm2gtfs/tests/creators/creators_tests.py
```diff
@@ -0,0 +1,368 @@
...
+                                )
+                            )
+                        return False
+                    else:
+                        # trip_point found with no difference, break for st2 loop
+                        # to continue for the next st1
+                        break
+            if not trip_point_found:
+                print(
+                    "No corresponing stop_time for trip_id={} and stop_sequence={}".format(
```
nlehuby: *corresponding.
Maybe `matching`

osm2gtfs/tests/creators/creators_tests.py
```diff
@@ -0,0 +1,368 @@
...
+        # Obtain and print basic count infos about the GTFS
+        gtfs_infos = CreatorsTestsHelper.get_gtfs_infos(gtfs_generated_result)
+        print("Information about the GTFS feed: " + str(gtfs_infos))
+
+        # Verify stop and route counts against expected values
+        self.assertEqual(gtfs_infos["stop_points_count"], self.required_values['stops_count'],
+                         "Wrong stop_points count in the generated GTFS")
+        self.assertEqual(gtfs_infos["stop_areas_count"], self.required_values['stations_count'],
+                         "Wrong stop_areas count in the generated GTFS")
+        self.assertEqual(gtfs_infos["routes_count"], self.required_values['routes_count'],
```
nlehuby: Can we switch the order of the tests ?
The tests about the required values provide much more useful info to debug and to inspect the potential differences than the `is_identical_gtfs` one.

osm2gtfs/tests/creators/creators_tests.py
```diff
@@ -0,0 +1,368 @@
...
+        * stops_count
+        * stations_count
+        * stops_osm_count
+        * route_id_to_check
+        * gtfs_files
+
+        """
+        raise NotImplementedError
+
+    def _get_standard_values(self):
```
nlehuby: I don't like `values`. Maybe `variables` or `test_config` ?

xamanu: Sure, I can rename it to `variables`.

osm2gtfs/tests/creators/creators_tests.py
xamanu: There has been no change on this one in this PR. But I'm happy to improve the wording as suggested.

osm2gtfs/tests/creators/creators_tests.py
xamanu: Of course, but probably better to do in a separate PR.

xamanu:
Thanks for your reviews @prhod and @nlehuby!

xamanu:
Rerolled with the latest suggestions. Further I included the query print-outs from #134, because it makes much more sense to have them in this PR.

From my side, this seems to be ready to merge in.

xamanu:
@grote Any chance to get this in at some point?

grote:
Thanks for the work and the reminder!

xamanu:
Thank you

PR #134, opened 2018-03-02
Tests for all providers

xamanu:
This PR is based on #133. It adds tests for all remaining existing providers:

* Florianopolis, Brazil (@grote)
* Esteli, Nicaragua (@ialokim, @AltNico)
* Gran Area Metropolitana, Costa Rica (@jamescr)


osm2gtfs/core/osm_connector.py
```diff
@@ -592,6 +592,7 @@ def _query_routes(self):
 
             /* Return tags for elements and roles for relation members. */
             );out body;""" % (self.tags, self.bbox)
+        print(query_str)
```
nlehuby: You forgot to remove the print here.

osm2gtfs/tests/creators/tests_br_florianopolis.py
```diff
@@ -0,0 +1,67 @@
...
+        with patch("osm2gtfs.core.osm_connector.OsmConnector._query_routes") as mocked1:
+            overpass_xml = open(mocked_overpass_data_file, mode='r').read()
+            api = overpy.Overpass()
+            mocked1.return_value = api.parse_xml(overpass_xml)
+            data.get_routes(refresh=True)
+        self.assertTrue(os.path.isfile(cache_file), 'The routes cache file creation failed')
+        cache = Cache()
+        routes = cache.read_data(self.selector + "-routes")
+        self.assertEqual(
+            len(routes), self.required_values['routes_count'] + 8,
```
nlehuby: Why do add 8 here ?

If it is really the right thing to do, I think it need a comment :thinking:

osm2gtfs/core/osm_connector.py
xamanu: I put this on purpose, in order to have the query printed as part of the general output.

osm2gtfs/tests/creators/tests_br_florianopolis.py
xamanu: It is because of how the Florianopolis creator works. I'm generally not happy with it. For now the "+8" made it work. But the whole creator needs a deeper revision in my opinion. But this can not be part of this PR.

I'm happy to hear suggestions to handle this better and on the wording of the comment.

osm2gtfs/core/osm_connector.py
xamanu: Moved over to PR #133.

osm2gtfs/tests/creators/tests_br_florianopolis.py
xamanu: I added a comment, as suggested by @nlehuby and moved the whole Florianopolis tests into an own PR: #135

xamanu:
I decided to split the tests for each provider into separate PRs, (#135, #136, #137) which can be reviewed more easily and comfortably by the respective creator specialists. And also allows to approve a certain area's tests while others are still getting improved or waiting for review. Closing here.

PR #135, opened 2018-03-05
Add acceptance tests for Florianopolis, Brazil

xamanu:
This PR is based on #134. It introduces a non-regression, acceptance tests for the Florianopolis provider. A review of the owner of the creator (@grote) is highly appreciated and a precondition to get this merged in.

This PR also includes a fix caused by [this commmit](https://github.com/grote/osm2gtfs/commit/3a1ece5be040d3f6f8699c83993e0c5f46fc3c5e), which itself is a good example, why those tests are relevant.

PR #136, opened 2018-03-05
Add acceptance tests for Esteli, Nicaragua

xamanu:
This PR is based on #134. It introduces a non-regression, acceptance tests for the Esteli provider. A review of at least one the owners of the creator (@ialokim and @AltNico) is highly appreciated and a precondition to get this merged in.


ialokim:
Not sure if I'm wrong, but isn't this testing exactly the same (standard) creator as the already accepted PR #131 ?

Instead of adding the same test for each one of the creators which use the standard creator's approach, wouldn't it be more straightforward and simple to only add a unit test for the standard creator once?

Of course it's important to add unit test for all creators that don't use the standard creator such as #137 and #135.

xamanu:
Good point! I don't have a strong opinion (yet). We could just have one test for a provider using the standard creators, or a test for each provider, and it doesn't matter, if it uses a custom or a standard creator. I see advantages and disadvantages for both approaches. Having tests for all, assures a bit more test coverage (as providers can still be different even when using the standard creators), but it is clearly adding up time to the tests to run.

xamanu:
Having an own test also allows providers to extend the class for particular tests. At the end, the more I think about it, the more I opt for having tests for each provider.

xamanu:
#129 is another use case where separate tests would be useful, wouldn't it?

xamanu:
Updated based on latest upstream. Now this PR only includes the respective commit and is easier to review, for your convenience.

xamanu:
Rebased on latest `master`.

grote:
This is awaiting review from @AltNico it seems.

grote:
Thanks @xamanu and @AltNico !

PR #138, opened 2018-04-03
Better warning message for unused stops in stop areas/stations

xamanu:
Unused members in stop areas/stations are common. Especially in the scenario when stations serve buses that are part of different GTFSes, like having one for a urban buses and one for interurban buses. Therefore this should be a "Note" rather than an "Error" and the wording can be more explanatory. Alternatively we could also consider to drop the message entirely.


grote:
Thanks @xamanu !

PR #139, opened 2018-04-03
Improve reporting of transformation of stop areas (OSM) to stations (GTFS)

xamanu:
When stop areas from OSM are transformed into GTFS stations, there are several different cases that are checked and reported. While in OSM a stop_area can (probably should) also be used to unite a `stop_position` and only one single `platform`, in GTFS the definition of stations requires several platforms to exist. Therefore a "Warning" should be changed to a "Note" and another "Note" is included for successful station creation.

### DIFF
--- a/osm2gtfs/core/osm_connector.py
+++ b/osm2gtfs/core/osm_connector.py
@@ -529,14 +529,14 @@ class OsmConnector(object):
             # group different elements of one stop together.
             sys.stderr.write(
                 "Error: Station with no members has been discarted:\n")
-            sys.stderr.write("https://osm.org/relation/" +
+            sys.stderr.write(" https://osm.org/relation/" +
                              str(stop_area.id) + "\n")
             return None
 
         elif len(members) == 1:
             sys.stderr.write(
-                "Warning: Station has only one platform and is discarted\n")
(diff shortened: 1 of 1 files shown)


osm2gtfs/core/osm_connector.py
```diff
@@ -520,31 +520,31 @@ def _build_station(self, stop_area, osm_type):
...
-            sys.stderr.write("https://osm.org/relation/" +
+            sys.stderr.write(" https://osm.org/relation/" +
                              str(stop_area.id) + "\n")
             return None
 
         elif len(members) == 1:
             sys.stderr.write(
-                "Warning: Station has only one platform and is discarted\n")
-            sys.stderr.write("https://osm.org/relation/" +
+                "Note: Stop area has only one platform and won't be a station \n")
```
nlehuby: This is not very easy to understand.
suggestion: `OSM stop area has only one platform and cannot be used as a GTFS station`

grote:
@xamanu This has a conflict. Would you mind resolving it and incorporating @nlehuby's suggestion? Thanks!

xamanu:
Yep. Done. Thanks for reviewing. Cheers.

Issue #46, opened 2016-11-19
Improve debugging


xamanu:
Debugging, such as screen output and special functions like `Route.print_shape_for_leaflet()` are currently within regular code. I'd like to optimize this. Right now I'm thinking about:

* a `Debug` class for special functions 
* making debug print output always depending on a flag (probably set through a bash argument)

xamanu:
There is some [ongoing work here](https://github.com/xamanu/osm2gtfs/tree/issue-46-debug). Still in progress.

grote:
Why not use the [Logger](https://docs.python.org/2/library/logging.html) class?

xamanu:
Good questions. Logger looks exactly like what I wanted to achieve. Let's go for this one.

Skippern:
I have user Logger on other projects, the only feature I miss is the ability to maintain multiple log files. Good thing is that dependencies using Logger will also be included in the log file, i.e. `requests`

xamanu:
I gave it a try with this PR: #142

xamanu:
This has been [merged in](https://github.com/grote/osm2gtfs/commit/210f65e8c7dfdaa82a8094ff5bac5e0259c1ee43). Thanks for the review @grote.

PR #142, opened 2018-04-03
Introduce python logging module

xamanu:
This is a PR related to issue #46 - Improve debugging by introducing python's [logging](https://docs.python.org/2/library/logging.html) module.

### DIFF
--- a/osm2gtfs/core/configuration.py
+++ b/osm2gtfs/core/configuration.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import logging
 import json
 import datetime
 from urllib2 import urlopen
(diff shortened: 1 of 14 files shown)


nlehuby:
There are still a few `print` that could be changed to logging messages in the following files:
* core/osm_connector.py
* creators/trips_creator.py

xamanu:
Thanks! Transformed the `print` into `logging.info` and rebased the PR on latest master.

xamanu:
It had again a little conflict after the latest PR has been accepted. Rebased again.

grote:
Do you have any idea why the CI is failing on this branch only?

xamanu:
No. I basically just rebased. Looks a bit like a strange error.

grote:
Could it be that the tests run a long time and now there's no more output where there was output before?

xamanu:
Yes, indeed. The logging level was not being set properly when running the tests, and fell back to the standard `WARNING`, which was not giving that regular output. Now, for tests the logging level `INFO` is set, too, and tests run successfully \o/

PR #143, opened 2018-08-23
add taginfo file to let OSMers know which tags we use

nlehuby:
Taginfo is a very useful project to learn about tags in OSM, and you can use it to know what the tags are used to.

this PR add a taginfo file for our project, so it can be displayed in the pages for the tags we use.

The documentation for the file format can be found here : https://wiki.openstreetmap.org/wiki/Taginfo/Projects

I didn't add too much info in here, because this does not have to replace our documentation, it is mainly a way to advertise our project and how to map in OSM to use it.

### DIFF
--- a/taginfo.json
+++ b/taginfo.json
@@ -1,0 +1,81 @@
+{
+    "data_format": 1,
+    "project": {
+        "name": "osm2gtfs",
+        "description": "Turn OSM public transport data and schedule information into GTFS",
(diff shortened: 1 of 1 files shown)


nlehuby:
Hello ?

grote:
Awesome, thank you for the PR! I didn't know about taginfo, but sounds useful indeed.

PR #144, opened 2018-10-13
Log missing stops and replace some sys prints with proper logging

grote:
Improve debugging output and fix typo
Downgrade schedule mismatch to non-critical error


grote:
Thanks for the quick review @xamanu!

PR #148, opened 2019-05-01
gh_accra - add links to the docs

nlehuby:


nlehuby:
thanks @grote and @prhod

PR #150, opened 2019-12-02
don't discard route without ref

nlehuby:
Our osm connector heavily rely on route/route_master `ref` tag.

Even if this tag is recommended, it can sometimes be missing in OSM without it being an error.
For instance, in Abidjan (Ivory Coast), we have wôrô-wôrô and gbakas that are artisanal transports operated in minibuses or repainted cars, which do not have any number or reference.

The OSM `ref` tag is usually used to set the GTFS `route_short_name`, which is an optional field, so we should not discard all route that do not have a `ref`.

This PR allows to use route/route_master without `ref` tag to create the route.txt file.
As it was used as a key to store the routes, I've updated the sorting of the routes in most creators so that the ids remain identical.

I will need your help to fix the last tests :wink:


osm2gtfs/core/osm_connector.py
```diff
@@ -576,6 +563,7 @@ def _query_routes(self):
         query_str = """(
             /* Obtain route variants based on tags and bounding box */
             relation%s(%s)->.routes;
+            //relation(10186002)->.routes;
```
xamanu: Seems like a debug line of code. Please remove.

osm2gtfs/core/osm_connector.py
```diff
@@ -170,14 +170,7 @@ def get_routes(self, refresh=False):
             if line is None:
                 continue
 
-            # Make sure route_id (ref) number is not already taken
-            if line.route_id in self.routes:
```
xamanu: I think this check for unique `ref` is still valid with the introduced changes. Probably should be moved to the creators now.

osm2gtfs/core/osm_connector.py
nlehuby: indeed, my bad

osm2gtfs/core/osm_connector.py
nlehuby: Hum, I disagree : I actually have two different bus lines ([Stigo 18](https://www.openstreetmap.org/relation/6957343) / [Seine-et-Marne Express 18](https://www.openstreetmap.org/relation/6957245)) with the same ref that run in my neighborhood. In Abidjan, we even have perfect duplicates, where one agency operates multiples lines with the same number (for instance the 13 one : https://www.openstreetmap.org/relation/10270789 vs https://www.openstreetmap.org/relation/10266011)
So if we want to keep this rules, it definitly should be in the creators and not in the core.

And I think we should handle duplicates in OSM instead of doing some magic to make them disappear in osm2gtfs.

But this is a different issue, so I can remove it from this PR if you prefer.

nlehuby:
I've investigated the Br_Florianopolis failing test:
The data have one route without ref (osm id = 7878890, line 47020 of the [overpass mock file](https://github.com/grote/osm2gtfs/blob/master/osm2gtfs/tests/creators/fixtures/br_florianopolis/overpass-routes.xml)).
This route is now present in the cache file (wasn't before) but is discarded in the creator.
I think this behaviour is ok. 
But I don't know what is the best fix: override the test data to remove this route vs changing the expected number of lines in the test. What do you think ?

osm2gtfs/core/osm_connector.py
xamanu: It looks like we have the common the idea, that the standard way of populating the GTFS `route_id` from OSM's `ref` tag - as this PR introduces. Highlighting here the [GTFS best practices](https://gtfs.org/best-practices/#routestxt) regarding the `route_id`:

> All trips on a given named route should reference the same route_id.
> * Different directions of a route should not be separated into different route_id values.
> * Different spans of operation of a route should not be separated into different route_id values. i.e. do not create different records in routes.txt for “Downtown AM” and “Downtown PM” services).

So, we have the situation that in OSM there can be duplicated `ref` values, as you line out in your comment, @nlehuby. And in GTFS they should be grouped to one entry. While I do think we have to build this bridge in osm2gtfs, I agree that the current logic of just ignoring the repeated occurrence is not addressing all use cases. But it is better than just dropping it entirely. 

I suggest we keep this logic in this PR, but move it to the creators. Then we can consider improving the logic in a separate issue, conversation and PR.

xamanu:
If the behaviour of dropping is desired, I would just lower the number of expected lines.

nlehuby:
Tests to fix:

- [ ] ni_esteli, AssertionError: Error found on stop_times for the route 3
- [ ] ni_managua, AssertionError: Error found on stop_times for the route 111
- [x] br_florianopolis, AssertionError: Wrong count of routes in the cache file

nlehuby:
It seems that the remaining failing tests are due to a change in the order of creation of the trips, that results in a difference in the trip_id.
But, there is an issue here: for now the trip are created in an impredictable order because we do not sort the dict (https://github.com/grote/osm2gtfs/blob/master/osm2gtfs/creators/trips_creator.py#L35)
For instance, for Esteli, we should add trips to feed in this order to make the tests pass: ['1', '3', '2', '5', '4'].

I think the good fix here is to sort the dict to have a predictable order and update the tests accordingly.
Would this be ok for you @xamanu or do you need to keep trip_id as they are at the present moment ?

xamanu:
> I think the good fix here is to sort the dict to have a predictable order and update the tests accordingly.
> Would this be ok for you @xamanu or do you need to keep trip_id as they are at the present moment ?

:+1: This would be an improvement for readability and therefor better control over the data.

osm2gtfs/core/osm_connector.py
nlehuby: > I suggest we keep this logic in this PR, but move it to the creators
it seems to me that no creators is ultimately using this logic. None of them have `ref` duplicates

nlehuby:
Thanks @xamanu.
This PR is now ready for review & merge.

xamanu:
Looks, generally good. But I would like to have the conversation about dropping or not the checks for duplicated 'ref' in a separate issue/PR instead of mixing it into this one here. Can you please bring the checks back (eventually in the standard creators)?

This allows us to proceed here quickly and get this in, while having a more detailed exchange of arguments around the checks, which I think is really necessary.

osm2gtfs/core/osm_connector.py
xamanu: Yes, because the data is good enough. The checks make sure that even if people add a duplicated 'ref' in OSM that osm2gtfs produces still a valid GTFS. As said, I think this can be improved, but taking it out, seems to be a step backwards and towards more inconsistent GTFS files.

xamanu: > And I think we should handle duplicates in OSM instead of doing some magic to make them disappear in osm2gtfs.

Yes, I deeply agree with that, and this is why I'm against dropping them.

In the case there is a duplicated `ref` in OSM data, but in GTFS there shall be only one unique `route_id`.

What happens with the checks:

* It generates a valid GTFS (omitting some routes, so incomplete).
* It displays an error message, so that people can:
  - Either fix it in OSM with one `route_master` for the different routes.
  - Use a custom creator to do their own logic.

What happens without the checks:

* It generates an invalid GTFS (with duplicated `route_id`)
* No warning message is presented to the user.

nlehuby:
ok, done

osm2gtfs/core/osm_connector.py
nlehuby: ok, I just understood your point.
I had forgotten that OSM `ref` tag was also used to create GTFS `route_id` (I usually use osm relation id as route_id) and I thought the debate was mostly about the `route_short_name` field. 
You are right, without the check, we have an invalid GTFS.

nlehuby:
Thanks @xamanu

PR #153, opened 2019-12-10
[WIP] Port to Python 3

pantierra:
Python 2 is reaching end of life. Preparing here to move to Python 3. Currently we depend on the `transitfeed` library becoming available in Python 3.


nlehuby:
Thanks, we really need to think about python3 now!

Why not have separate PR ? We can start using flake8 and black even if we are not python3 ready yet, can't we ?

xamanu:
In theory, yes. But unfortunately we have already [problems installing black in the CI with Python 2](https://travis-ci.org/grote/osm2gtfs/jobs/644727482?utm_medium=notification&utm_source=github_status). We could still do the black improvements, but without the checks...

nlehuby:
`transitfeed` migration to python3 is still not done and it doesn't look like it's really moving forward. Any idea how we can work this out on our end?

pantierra:
Superseeded by #160

Issue #146, opened 2019-01-04
Default implementation for TripCreator with frequency support


nlehuby:
We should allow the default TripCreator to work without schedule data and then generate a GTFS with a `frequency.txt` file that only uses OSM data.

Even if it won't fit all needs, it will allow to create very easily a first GTFS from OSM data that could even be usable for some simple cases (long distances buses with low frequency, ferries and shuttle bus, subways with high frequency, etc)

We can use 
* the `interval` OSM tag to get the frequency : https://wiki.openstreetmap.org/wiki/Key:interval
* the `opening_hours` to define the stop time at the first stops
* the `duration` key (and the GTFS interpolate stop time utility) to deduce times at each stops

The [Accra creator](https://github.com/grote/osm2gtfs/blob/master/osm2gtfs/creators/gh_accra/trips_creator_gh_accra.py) is a good start to implement this as it already does that. 
Well, almost :
* it has its `opening_hours` hard coded
* it uses `travel_time` tag instead of `duration` and `frequency` tag instead of `interval`
* the tags are read on either the route or the route_master (we should to be smart enough to handle both)
* the conditional part of the `interval` tag is not supported

xamanu:
Currently the standard trips creator uses a kind of [standartized time table input format](https://github.com/grote/osm2gtfs/wiki/Schedule), with the idea to cover all possibilities. In Managua we also have frequency/interval based data, we just converted it to the timetable format with a simple script. Your proposed direct interval approach is surely a simplification, which is interesting.

Is your proposal to remove the timetable input and to replace it with the `interval` approach? Or do think on supporting both? In the latter case, how do you suggest to differentiate between the two?

ialokim:
> converted it to the timetable format with a simple script

Just for reference, you can find it [here](mapanica/easy-timetable-generator).
I think back then we went for the standard `stop_times.txt` approach as the `frequencies.txt` information seemed to be not fully supported by some GTFS consumers (?). There was also the idea of integrating the script into the `osm2gtfs` code.

> to differentiate between the two [approaches]?

I would say, if the `schedule_source` key is present and set within the configuration (as e.g. [for Estelí](grote/osm2gtfs/blob/master/osm2gtfs/creators/ni_esteli/config.json#L32)), it would use the already available standard approach. If it is not present, `osm2gtfs` could automatically use the schedule information from OSM or dump default values if there is no information on OSM either as proposed in #118.

nlehuby:
[...]

PR #154, opened 2019-12-13
Abidjan Creator

nlehuby:
Here come the first version of the Abidjan (Ivory Coast) creator :tada:

It uses `operator` and `operator:website` tags from OSM relations to set up the GTFS `agencies`.

It generates a GTFS with a frequencies.txt file and uses the logic described in #146:
- the `opening_hours`, `interval` and `interval:conditional` tags on OSM relations are used to set up the schedule data
- the `duration` tag is used to interpolate the time of travel between stops of each trip

We will use it in Abidjan for a while (our data collection is still ongoing) and if everything works as expected, we will go on with the issues #146 and #118, and move this logic to the default creator.

Some tests comes along the creator.


nlehuby:
Hey ! any thought about this PR ?

Merry Christmas to all the bus mappers in the world :santa:  :busstop: :world_map:

xamanu:
Looks good! I didn't look too much into the custom trips creator. Are you finished? or are you still working on it?

Can you squash the commits to a single one? Maybe with the two of you, one with `Co-authored-by:`.

nlehuby:
Yes, we are finished here. You can use the "squash and merge" github button

pantierra (approved):

PR #137, opened 2018-03-05
Add acceptance tests for Gran Area Metropolitana, Costa Rica

xamanu:
This PR is based on #134. It introduces a non-regression, acceptance tests for the Costa Rican provider. A review of the owner of the creator (@jamescr) is highly appreciated and a precondition to get this merged in.

PR #158, opened 2020-01-18
Standard creators for incofer

jamescr:
Incofer schedule data was adapted to the osm2gtfs schedule structure. Now there's no need for the incofer custom creators, the gtfs is created using the standard ones.


nlehuby:
That's pretty cool @jamescr :)

Can you please add regression tests so we can make sure we don't break it in the future without knowing ?
See #75 and #137

jamescr:
> That's pretty cool @jamescr :)

Thanks

> Can you please add regression tests so we can make sure we don't break it in the future without knowing ?
> See #75 and #137

I have already made [another PR](https://github.com/xamanu/osm2gtfs/pull/2) to @xamanu with an update of the mock data needed for the tests but assuming this PR gets merged.

pantierra (approved):
This is great!

Issue #156, opened 2019-12-30
Support hail_and_ride / GTFS Flex


amenk:
Are the plans to support GTFS flex and hail_and_ride tagging?

nlehuby:
I think it would be a good addition to osm2gtfs :+1:

But IMHY we should wait for it is fully integrated into the official GTFS spec.

For the record :
* [the GTFS extention proposal](https://docs.google.com/document/d/1QVrijt2omNN_lvCkC0FmRhciVmWg-cJtNW_PqULc-Pw/edit#heading=h.9i1fb17t3lbp)
* [the OSM proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/hail_and_ride) (that has been approved in 2018)

Skippern:
What is meant by hail_and_ride? Fixed itinerary without fixed stop positions or flexible itinerary? I have seen both solutions, rural bus services in Norway are generally fixed itinerary without fixed stop positions, while some minor places have "taxibus" services where you call a number or via app, and your location + destination is added to current itinerary.

amenk:
It's about Addis Ababa minibus service. The routes are fixed, but they let you hop on and off anywhere on that route

amenk:
> Does OTP support GTFS flex in any way, because that is what we e generating for..

http://docs.opentripplanner.org/en/latest/Flex/

amenk:
For minibus in Addis Ababa it would be 

https://github.com/MobilityData/gtfs-flex/blob/master/spec/reference.md#request-stops

> 3 - Must coordinate with driver to arrange continuous stopping behavior.

continuous_pickup=3
continuous_drop_off=3 

in stop_times.txt

amenk:
We might need to fork transitfeed, because it does not support that fields 

https://github.com/google/transitfeed/blob/master/transitfeed/stoptime.py#L48

amenk:
I made some tests with OTP and manually adding to stoptimes.txt

	"trip_id","arrival_time","departure_time","stop_id","stop_sequence","stop_headsign","pickup_type","drop_off_type","shape_dist_traveled","timepoint","continuous_pickup","continuous_drop_off"
	24,"06:00:00","06:00:00","node/7037142417",1,,,,,,0,0
	24,"06:04:54","06:04:54","node/7110438109",2,,,,,,0,0

....

And router-config.jsom

	{
	  "useFlexService": true,
	  "routingDefaults": {
	    "flexCallAndRideReluctance": 3,
[...]

PR #159, opened 2020-02-02
Add Addis Ababa and Add Partial GTFSFlex Implementation (#156)

amenk:
* Mostly adding Addis Ababa Scripts
* Also add transitfeedflex Module to support Minibus System in Addis Ababa


osm2gtfs/creators/et_addisababa/config.json
```diff
@@ -0,0 +1,35 @@
+{
```
nlehuby, line 1: All config should come with some regression tests, so we can make sure our future changes in osm2gtfs do not break your own implementation.

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
```diff
@@ -0,0 +1,133 @@
...
+            print("Generating schedule for line: " + route_ref)
+
+            flex_flag = None
+            if 'route_master' in line.tags and line.tags['route_master'] == "light_rail":
+                route_type = "Tram"
+                route_suffix = " (Light Rail)"
+            elif 'route_master' in line.tags and line.tags['route_master'] == "share_taxi":
+                route_type = "Bus"
+                route_suffix = " (Minibus)"
+                flex_flag = 0
```
nlehuby, line 40: if `flex_flag` is a flag, can we use a Boolean ?

nlehuby, line 40: I don't like the idea of it being hard-coded. Can we rather have that information as an attribute on the route / route_master in OSM ?

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
```diff
@@ -0,0 +1,133 @@
...
+
+        lines = data.routes
+        for route_ref, line in sorted(lines.iteritems()):
+            if not isinstance(line, Line):
+                continue
+            print("Generating schedule for line: " + route_ref)
+
+            flex_flag = None
+            if 'route_master' in line.tags and line.tags['route_master'] == "light_rail":
+                route_type = "Tram"
```
nlehuby, line 35: I don't understand why you need to override the route_type here :thinking:

transitfeedflex/README
```diff
@@ -0,0 +1 @@
+Adds Flexible Pick Off / Drop Off (Part of the GTFSFlex specification)
```
nlehuby, line 1: :+1: That's really cool to add this feature!

Can you please add a link to the documentation of the format ?

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
```diff
@@ -0,0 +1,133 @@
...
+                            print("frequency is invalid for route_master " + str(
+                                line.osm_id))
+                    except (ValueError, TypeError) as e:
+                        print("frequency not a number for route_master " + str(
+                                line.osm_id))
+
+                trip_gtfs.AddFrequency(
+                    "05:00:00", "22:00:00", ROUTE_FREQUENCY * 60)
+
+                if 'duration' in a_route.tags:
```
nlehuby, line 94: the `ci_abidjan` creator also use `duration` and `interval` tags for the same purpose :thinking: I wonder if / how we can mutualize the implementations to avoid code duplication.

nlehuby (commented):
Nice work @amenk

transitfeedflex/README
amenk: long time no feedback, sorry :-)
I think it is described here: https://github.com/MobilityData/gtfs-flex/blob/master/spec/reference.md

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
amenk: In OSM we tagged it as light_rail, and I was thinking it needs to be mapped to Tram in GTFS; but I am not sure

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
amenk: Possible, need to discuss with the OSM community how it is mapped best on OSM side.

amenk: Probably yes :-)

osm2gtfs/creators/et_addisababa/config.json
amenk: I will probably not be able to add those from my current limited python skills. Help is appreciated.

osm2gtfs/creators/et_addisababa/trips_creator_et_addisababa.py
amenk: Interesting, how could we do this?

amenk:
Long time no see - sorry, I was kind of pushed back by the requirements for regression tests and also we did not work for some while on the project.

We want to bring this pull request forward :-) Unfortunately I have to confess that I only limited Python skills, so might need some support. But we will try our best to get this ready for merge.

amenk:
Actually I am nowadays no longer sure if OpenTripplanner v2 (which we use for routing) supports those continuous_pickup + continuous_drop_off fields.

It's also not really documented any more in the spec.

Ref: https://github.com/MobilityData/gtfs-flex/issues/70
https://github.com/opentripplanner/OpenTripPlanner/pull/2603

derhuerst:
> Actually I am nowadays no longer sure if OpenTripplanner v2 (which we use for routing) supports those continuous_pickup + continuous_drop_off fields.

For better discoverability, this is [Leonard's reply on the OTP mailing list](https://groups.google.com/g/opentripplanner-users/c/A9KQMtWhYBI/m/7mcEXn33BAAJ):

> continuous_pickup/dropoff is currently not supported by OTP2.
> There is however an open PR that would need some polishing to get merged: https://github.com/opentripplanner/OpenTripPlanner/pull/3457
> Several people have expressed interest in taking over from the original author but so far no-one has stepped forward.

(Sorry if this is irrelevant, I'm a random lurker on this Issue.)

hannesj:
> It's also not really documented any more in the spec.

It is part of the main GTFS specification https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#stop_timestxt

